PR #10, opened 2019-07-19
Fault codes client

richardiphillips:


src/Service.Host/client/src/components/ate/AteFaultCode.js
```diff
@@ -0,0 +1,186 @@
...
+                                disabled={!creating()}
+                                value={ateFaultCode.faultCode}
+                                label="Sequence Name"
+                                maxLength={10}
+                                helperText={
+                                    !creating()
+                                        ? 'This field cannot be changed'
+                                        : `${faultCodeInvalid() ? 'This field is required' : ''}`
+                                }
+                                error={faultCodeInvalid()}
```
atcampbell: There's now a required option for input fields which styles them with a red asterisk and red 'required' text. I think that's better for indicating that something is empty and keeping error state for invalid input etc

src/Service.Host/client/src/components/ate/AteFaultCode.js
```diff
@@ -0,0 +1,186 @@
...
+                            />
+                        </Grid>
+                        <Grid item xs={8}>
+                            <InputField
+                                value={ateFaultCode.description}
+                                label="Description"
+                                maxLength={50}
+                                fullWidth
+                                helperText={descriptionInvalid() ? 'This field is required' : ''}
+                                error={descriptionInvalid()}
```
atcampbell: SAme here with the required prop

richardiphillips:
Added two different types of tests. Found that I quite liked the old style for simple display screens where all I really care is that the table is displayed but like the new one for more intricate screens like the create / update page. 

Is this wrong?

PR #30, opened 2019-07-23
Outstanding works order report

atcampbell:


src/Service.Host/client/src/components/reports/OutstandingWorksOrdersReport.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { Grid } from '@material-ui/core';
```
lewisrenfrew: probs best to import Grid from '@material-ui/core/Grid' to prevent the whole lib being pulled in

PR #54, opened 2019-07-25
Builds by department report

lewisrenfrew:
- everything works as it should here but enumerating over the Queryable takes forever. I'll probably just rewrite this with the sql hardcoded, but I thought I would pr it in this state to see if anyone had any suggestions.


src/Persistence.LinnApps/Repositories/BuildsRepository.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            this.serviceDbContext = serviceDbContext;
+            this.lrpPack = lrpPack;
+            this.weekPack = weekPack;
+        }
+
+        public IEnumerable<BuildsSummary> GetBuildsByDepartment(DateTime from, DateTime to, bool monthly = false)
+        {
+            return this.serviceDbContext.Builds
+                .Where(b => b.BuildDate > from && b.BuildDate < to).ToList()
```
lewisrenfrew: the offending query

src/Proxy/DatabaseService.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Linn.Production.Proxy
+{
+    using System.Data;
+
+    using Oracle.ManagedDataAccess.Client;
+
+    public class DatabaseService : IDatabaseService
```
andrewfraser73: is this going to lead to a new oracle connection per request or do they stay open for longer - even if you close them when you're done there is an overhead in establishing and closing a connection each time

could just share one oracle connection for the entire solution

src/Service.Host/.vscode/settings.json
```diff
@@ -0,0 +1,23 @@
+{
```
atcampbell: This should be deleted - I think I added /vscode into the gitignore so might be fine after a merge.

src/Proxy/LRPPack.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            this.db = db;
+        }
+
+        public decimal GetDaysToBuildPart(string partNumber, decimal quantity)
+        {
+            using (var connection = this.db.GetConnection())
+            {
+                connection.Open();
+                var cmd = new OracleCommand("LRP_PACK.Days_To_Build_Part", connection)
```
andrewfraser73: if we keep repeating this pattern of calling oracle functions or procedures we could make some helper code to avoid the repetition

src/Service.Host/client/src/components/buildsbyDepartment/BuildsSummaryReportOptions.js
```diff
@@ -0,0 +1,69 @@
+import React, { useState } from 'react';
+import { Button, Grid, Typography } from '@material-ui/core';
+import { OnOffSwitch } from '@linn-it/linn-form-components-library';
+import { DatePicker } from '@material-ui/pickers';
+import PropTypes from 'prop-types';
+import Page from '../../containers/Page';
+
+function BuildsByDepartmentReportOptions({ history }) {
+    const [fromDate, setFromDate] = useState(new Date('2006-01-27'));
```
atcampbell: Why these default dates?

src/Service.Host/client/src/components/buildsbyDepartment/BuildsSummaryReportOptions.js
```diff
@@ -0,0 +1,69 @@
...
+    return (
+        <Page>
+            <Grid style={{ marginTop: 40 }} container spacing={3} justify="center">
+                <Grid item xs={12}>
+                    <Typography variant="h6" gutterBottom>
+                        Choose a date range:
+                    </Typography>
+                </Grid>
+                <Grid item xs={3}>
+                    <DatePicker label="From Date" value={fromDate} onChange={setFromDate} />
```
atcampbell: Do you think we should add this into the shared components or is it ok calling it like this?

src/Service.Host/client/src/containers/initialiseOnMount.js
```diff
@@ -0,0 +1,17 @@
+import React, { useEffect } from 'react';
```
atcampbell: Should this be removed and use the library one?

src/Service.Host/client/src/index.js
```diff
@@ -1,6 +1,8 @@
 ﻿import React from 'react';
 import ReactDOM from 'react-dom';
 import { AppContainer } from 'react-hot-loader';
+import { MuiPickersUtilsProvider } from '@material-ui/pickers';
+import MomentUtils from '@date-io/moment';
```
atcampbell: What are these doing?

lewisrenfrew: It's just a dependency injection way of telling the pickers which date library to use (in this case moment)

src/Service.Host/client/src/components/buildsbyDepartment/BuildsSummaryReportOptions.js
lewisrenfrew: because they are the two greatest days in history

src/Service.Host/client/src/components/buildsbyDepartment/BuildsSummaryReportOptions.js
lewisrenfrew: I think we should probably add it eventually just so it's easier to swap out if we ever go for a different date picker... but its so simple to use at the moment that I didn't really see the point adding it to the shared library

src/Service.Host/client/src/containers/initialiseOnMount.js
lewisrenfrew: this is slightly different to the shared library one. I need to test whether it breaks anything before making it the de facto initialiseOnMount

PR #58, opened 2019-07-29
Production measures backend

andrewfraser73:
Backend for production measures (OSR)


src/Resources/ProductionMeasuresResource.cs
```diff
@@ -0,0 +1,75 @@
...
+        public double? StockValue { get; set; }
+
+        public double? OverStockValue { get; set; }
+
+        public int? NumberOfBackOrders { get; set; }
+
+        public int? NumberOfPartsBackOrdered { get; set; }
+
+        public DateTime? OldestBackOrder { get; set; }
+
```
lewisrenfrew: I normally keep dates as iso strings in the resource... not sure if that's correct though

andrewfraser73: No I think that was our position looking at products maint

PR #60, opened 2019-07-31
Builds detail report

lewisrenfrew:
- #52  provides drilldown for the builds summaries report

### DIFF
--- a/src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
+++ b/src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
@@ -1,0 +1,73 @@
+﻿using System.Linq;
+using Linn.Common.Persistence;
+using Linn.Production.Domain.LinnApps.Reports;
+
+namespace Linn.Production.Domain.LinnApps.Services
+{
+    using System;
+    using System.Collections.Generic;
+    using Common.Reporting.Models;
+
+
+    public class BuildsDetailReportService : IBuildsDetailReportService
+    {
+        private readonly IBuildsDetailReportDatabaseService databaseService;
+
+        private readonly IRepository<Department, string> departmentRepository;
+
+        public BuildsDetailReportService(
+            IRepository<Department, string> departmentRepository,
+            IBuildsDetailReportDatabaseService databaseService)
+        {
+            this.databaseService = databaseService;
+            this.departmentRepository = departmentRepository;
+        }
+
+        public ResultsModel GetBuildsDetailReport(DateTime from, DateTime to, string department, string quantityOrValue,
+            bool monthly = false)
+        {
+            var table = databaseService.GetBuildsDetail(from, to, quantityOrValue, department, monthly);
+            var partGroups = table.Select().GroupBy(r => r[2]).ToList(); 
+            var weeks = partGroups.Select(g => ((DateTime)g.First().ItemArray[4]).ToShortDateString()).Distinct().ToList();
+
+            var colHeaders = new List<string> {"Part Number"};
+            colHeaders.AddRange(weeks);
+            colHeaders.Add("Total");
+            var results = new ResultsModel(colHeaders)
+            {
+                ReportTitle = new NameModel($"Builds {quantityOrValue} for {departmentRepository.FindById(department).Description}")
+            };
+            var rowIndex = 0;
+            foreach (var partGroup in partGroups)
+            {
+                var partTotal =  new decimal();
+                results.AddRow(partGroup.Key.ToString());
+                results.SetGridTextValue(rowIndex, 0, partGroup.Key.ToString());
+                
+                for (var i = 0; i < weeks.Count; i++)
+                {
+                    var valueExistsThisWeek = partGroup.FirstOrDefault(g =>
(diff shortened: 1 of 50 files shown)


src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.databaseService = databaseService;
+            this.departmentRepository = departmentRepository;
+        }
+
+        public ResultsModel GetBuildsDetailReport(DateTime from, DateTime to, string department, string quantityOrValue,
+            bool monthly = false)
+        {
+            var table = databaseService.GetBuildsDetail(from, to, quantityOrValue, department, monthly);
+            var partGroups = table.Select().GroupBy(r => r[2]).ToList(); 
+            var weeks = partGroups.Select(g => ((DateTime)g.First().ItemArray[4]).ToShortDateString()).Distinct().ToList();
```
atcampbell: Is it dangerous to rely on the [2] and [4] indexes being hard coded here?

src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
```diff
@@ -0,0 +1,73 @@
...
+                ReportTitle = new NameModel($"Builds {quantityOrValue} for {departmentRepository.FindById(department).Description}")
+            };
+            var rowIndex = 0;
+            foreach (var partGroup in partGroups)
+            {
+                var partTotal =  new decimal();
+                results.AddRow(partGroup.Key.ToString());
+                results.SetGridTextValue(rowIndex, 0, partGroup.Key.ToString());
+                
+                for (var i = 0; i < weeks.Count; i++)
```
atcampbell: Could this be changed to another foreach or a linq expression `weeks.Select(week => blah)` ? Then you might not neet Weeks.elementAt(i)?

src/Service.Host/client/src/components/buildsbyDepartment/BuildsDetailReportOptions.js
```diff
@@ -0,0 +1,106 @@
...
+                <Grid item xs={12}>
+                    <Typography variant="h6" gutterBottom>
+                        Report Options:
+                    </Typography>
+                </Grid>
+                <Grid item xs={6}>
+                    <Dropdown
+                        label="Department"
+                        propertyName="department"
+                        items={[...departments.map(d => d.description)]}
```
atcampbell: Would this work without the [... ], map returns an array anyway?

src/Service.Host/client/src/containers/buildsByDepartment/BuildsDetailReport.js
```diff
@@ -0,0 +1,40 @@
...
+const reportSelectors = new ReportSelectors('buildsDetailReport');
+
+const getOptions = ownProps => {
+    const options = queryString.parse(ownProps.location.search);
+    if (options.fromDate === 'undefined') {
+        options.fromDate = new Date().toISOString();
+    }
+    if (options.toDate === 'undefined') {
+        options.toDate = new Date().toISOString();
+    }
```
atcampbell: Would these be better as selectors so they can be tested etc?

src/Service.Host/client/src/components/buildsbyDepartment/BuildsDetailReportOptions.js
lewisrenfrew: yeah... I have no idea why I typed a ... there.

src/Service.Host/client/src/containers/buildsByDepartment/BuildsDetailReport.js
lewisrenfrew: I'm not sure... I guess they're coming from props and not state as is the usual case for a redux selector, i.e.

`selector = state => do something with state`.

I'm probably being overly defensive here anyway as it's highly unlikely that someone could get to this page with out the dates in the query string... and it doesn't really make sense to just pass in a new Date() in that case anyway... Not sure what I was thinking ¯\_(ツ)_/¯

src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
lewisrenfrew: a little... if the query changes these'll fall out of sync. Not sure how else to access specific values in the datagrid returned by oracle queries though... Can you do it by column name or something as opposed to number?

src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
lewisrenfrew: I went for the classic for loop because I knew I'd need  to know i + 1 in the loop. Can you think of a more succinct way to do this stuff?

atcampbell: Add I didn't notice the i + 1. I think you could do `weeks.select((week, index) => blarg)` maybe not though...

src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
atcampbell: Yeah I have no idea actually... 😢

PR #66, opened 2019-08-08
OSR initial page

andrewfraser73:
This is the first draft of the OSR (Operations Status Report) none of the drill downs are in place.  Also the initial report contains other data around the run times and ATE and board pass rates

Have also written a CSV export

The main UI is an amalgamation of tabs and table.


src/Facade/ResourceBuilders/ProductionMeasuresResourceBuilder.cs
```diff
@@ -27,7 +27,7 @@ public ProductionMeasuresResource Build(ProductionMeasures productionMeasures)
                 StockValue = productionMeasures.StockValue,
                 BackOrderValue = productionMeasures.BackOrderValue,
                 PboJobId = productionMeasures.PboJobId,
-                OldestBackOrder = productionMeasures.OldestBackOrder?.ToString("o"),
+                OldestBackOrder = productionMeasures.OldestBackOrder?.ToString("dd-MMM-yy"),
```
atcampbell: Why did you need to change this from an iso string?

src/Service.Host/client/src/components/reports/measures/ProductionMeasuresCitTable.js
```diff
@@ -0,0 +1,363 @@
...
+    const shortagesBaseHref = `${config.appRoot}/production/reports/measures/shortages`;
+    const backOrdersBaseHref = `${config.appRoot}/production/reports/measures/back-orders`;
+    const builtThisWeekBaseHref = `${config.appRoot}/production/reports/measures/built-this-week`;
+    const fflagStockBaseHref = `${config.appRoot}/production/reports/stock/f-flag`; // may already exist in linnapps
+
+    function handleChange(event, newValue) {
+        setValue(newValue);
+    }
+
+    function round(num) {
```
atcampbell: Is it worth adding this to the shared library?

src/Service.Host/client/src/components/reports/measures/ProductionMeasuresCitTable.js
```diff
@@ -0,0 +1,363 @@
...
+                        <TabCell index={0} value={tabValue} />
+                        <TabCell index={0} value={tabValue} />
+                        <TabCell index={0} value={tabValue} />
+                        <TabCell index={0} value={tabValue} align="centre" colSpan={2}>
+                            Days
+                        </TabCell>
+                        <TabCell index={0} value={tabValue} align="centre" colSpan={2}>
+                            Can Do
+                        </TabCell>
+                        <TabCell index={0} value={tabValue} />
```
atcampbell: What are these value={tabValue} doing?

src/Facade/ResourceBuilders/ProductionMeasuresResourceBuilder.cs
andrewfraser73: it was what was displayed on the OSR report converting it to a string with full ISO pattern would then require further work downstream when this is literally the only place it will be consumed

src/Service.Host/client/src/components/reports/measures/ProductionMeasuresCitTable.js
andrewfraser73: maybe certainly if it is used on another component there is a case for moving it

src/Service.Host/client/src/components/reports/measures/ProductionMeasuresCitTable.js
andrewfraser73: maybe value isn't the best description it is the index of the currently selected tab hence tabValue.  i will ponder on whether there is a better name for this.

Issue #33, opened 2019-07-23
PTest Board Fail Types Utility


andrewfraser73:
Ptest and board test history
Board Fail Types
BOARD_FAIL_TYPE_UT
Maintenance screen

PR #68, opened 2019-08-09
Fail types utility #33

lewisrenfrew:


src/Domain.LinnApps/Reports/BuildsSummaryReportService.cs
```diff
@@ -32,7 +32,7 @@ public IEnumerable<ResultsModel> GetBuildsSummaryReports(DateTime from, DateTime
                                                    })
                                   {
                                       ReportTitle = new NameModel(
-                                          week.Key.ToString("d", CultureInfo.CurrentCulture))
+                                          week.Key.Date.ToString("d", new CultureInfo("en-GB")))
```
atcampbell: Should it be "o" for an iso string here?

src/Service.Host/client/src/components/boardFailTypes/BoardFailTypes.js
```diff
@@ -0,0 +1,131 @@
...
+        orderAscending: false,
+        currentPage: 0,
+        rowsPerPage: 10
+    });
+    const [rowsToDisplay, setRowsToDisplay] = useState([]);
+
+    const classes = useStyles();
+
+    useEffect(() => {
+        const compare = (field, orderAscending) => (a, b) => {
```
atcampbell: I've put a couple of compare functions in the shared library that might work here

lewisrenfrew: nice, but I actually just axed the sorting capability as i didn't need it and blindly copied and pasted this code

src/Domain.LinnApps/Reports/BuildsSummaryReportService.cs
lewisrenfrew: nah, this is just formatting it for display on the report, so just going with uk date format

Issue #25, opened 2019-07-23
Assembly Fails Waiting List


andrewfraser73:
Internal Quality Measures
Assembly Quality System
Waiting List
ASSEMBLY_FAILS_R1

PR #73, opened 2019-08-14
Assembly fails waiting list #25

lewisrenfrew:


lewisrenfrew:
> Looks good. If it does everything the other one does then we should start finding who uses it, perhaps from the usage stats, and getting them to try/use it.
It should drill down the assembly fails utility, so gonna do that next and then I'll find some users

PR #77, opened 2019-08-23
back end for osr info (smallish)

andrewfraser73:
Back end to get PTL master and OSR run info so I can display some basic info on the OSR page and have the right PTL jobref to pass to the PTL drilldown (coming soon).

introduces a new cut down repo interface for master tables which have one record only.  if proves useful could move into common.

osrinfo will expand with other info required for the report to avoid multiple HTTP calls to the backend when one can do all the SQL required


src/IoC/PersistenceModule.cs
```diff
@@ -5,6 +5,7 @@
     using Domain.LinnApps.ATE;
     using Domain.LinnApps.Measures;
     using Domain.LinnApps.SerialNumberReissue;
+    using Domain.LinnApps.Triggers;
```
richardiphillips: why does your pc miss out Linn.Production. no matter how many times it gets changed? 😃

src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -1,4 +1,7 @@
-﻿namespace Linn.Production.Persistence.LinnApps
+﻿using System.Linq;
```
richardiphillips: should be in namespace

src/IoC/PersistenceModule.cs
andrewfraser73: i make sure Kenny always transfers over my user state so it keeps my ridiculous settings.  However I tracked down my resharper setting to prefer fully qualified usings and fixed these

PR #82, opened 2019-08-26
Osr Info UI

andrewfraser73:
Feeding the OSR info through to the UI
- last trigger jobref passed through to drilldowns
- Info button to show last run and assorted semi useful info
- Potential new shared component InfoButton using Material UI Popover
- changes to tests

Only 15 files to change


src/Service.Host/client/src/reducers/productionMeasures/info.js
```diff
@@ -0,0 +1,22 @@
+function productionMeasuresCits(state = { loading: false, data: null }, action) {
+    switch (action.type) {
+        case 'REQUEST_PRODUCTION_MEASURES_INFO_REPORT':
+            return {
+                ...state,
+                loading: true,
```
richardiphillips: Can't you use the standard report reducers and selectors even if you aren't using the standard components?

andrewfraser73: the standard reducers and selectors assume you are using a reportData return format and error if you aren't (I did try this for a few hours).  We have generated solutions for full CRUD and for reportData but this is somewhere in the middle.  When I do the drill downs I will try to use reportData if I can.

PR #85, opened 2019-08-29
Works order report options

atcampbell:


src/Facade/ResourceBuilders/CitResourceBuilder.cs
```diff
@@ -0,0 +1,37 @@
...
+    using Linn.Common.Facade;
+    using Linn.Common.Resources;
+    using Linn.Production.Domain.LinnApps.Measures;
+    using Linn.Production.Resources;
+
+    public class CitResourceBuilder : IResourceBuilder<Cit>
+    {
+        public CitResource Build(Cit cit)
+        {
+            return new CitResource()
```
richardiphillips: don't need ()

src/Service.Host/client/src/components/reports/OutstandingWorksOrdersReportOptions.js
```diff
@@ -0,0 +1,90 @@
+import React, { Fragment, useState, useEffect } from 'react';
+import { Loading, Title, Dropdown } from '@linn-it/linn-form-components-library';
+import { Button, Grid } from '@material-ui/core';
```
richardiphillips: I think Lewis is a proponent of always bringing these in individually

PR #87, opened 2019-08-29
PTL Backend

andrewfraser73:
Backend for production trigger report (and groundwork for the why-oh-why trigger drilldown)
- mimics shape of reportresults
- uses Dbquery instead of dbset
- incubates IQueryRepository a cut down IRepository for when you don't have a key field
- new oracle view V_production_triggers_Ef does some of the heavy lifting and checks in SQL especially around ordering and Brief/Full reporting options


src/Domain.LinnApps/Triggers/ProductionTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public decimal? NettSalesOrders { get; set; }
+
+        public int? QtyFree { get; set; }
+
+        public int? RemainingBuild { get; set; }
+
+        public int? QtyBeingBuilt { get; set; }
+
+        public int? BE { get; set; }
```
richardiphillips: I know we're staying true to the oracle field names and I think that's right but I wonder if we should break that rule for these particularly obscure names. They're torturous for us and unintelligible for other people. 
I know I curse the sos ones every time I have to re-interpret them

src/Facade/Services/ProductionTriggersFacadeService.cs
```diff
@@ -0,0 +1,77 @@
...
+            var cit = this.citRepository.FindById(citCode);
+
+            if (cit == null)
+            {
+                return new NotFoundResult<ProductionTriggersReport>($"cit {citCode} not found");
+            }
+
+            var ptlMaster = masterRepository.GetMasterRecord();
+
+            var triggers = (triggerReportType == ProductionTriggerReportType.Full)
```
richardiphillips: I think all this logic to build the actual report should be in the domain rather than the facade service which should just wrap it with a success/failure result type.

src/Facade/Services/ProductionTriggersFacadeService.cs
```diff
@@ -0,0 +1,77 @@
...
+                ? repository.FilterBy(t => t.Jobref == jobref && t.Citcode == citCode)
+                : repository.FilterBy(t => t.Jobref == jobref && t.Citcode == citCode && t.ReportType == "BRIEF");
+
+            triggers = triggers.OrderBy(t => t.SortOrder).ThenBy(t => t.EarliestRequestedDate);
+
+            var report = new ProductionTriggersReport()
```
richardiphillips: don't need ()

src/IoC/ResponsesModule.cs
```diff
@@ -11,6 +11,8 @@
     using Linn.Production.Domain.LinnApps;
     using Linn.Production.Domain.LinnApps.ATE;
     using Linn.Production.Domain.LinnApps.Measures;
+    using Domain.LinnApps.SerialNumberReissue;
```
richardiphillips: where did this appear from? No Linn.Production

tests/Unit/Facade.Tests/ProductionTriggersFacadeServiceSpecs/WhenGettingTriggersWithNoJobRef.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Linn.Production.Facade.Tests.ProductionTriggersFacadeServiceSpecs
+{
+    using FluentAssertions;
+    using Linn.Common.Facade;
+    using Linn.Production.Domain.LinnApps.Triggers;
+    using NUnit.Framework;
+
+    public class WhenGettingTriggersWithNoJobRef : ContextBase
+    {
+
```
richardiphillips: blank line in all the tests.

tests/Unit/Facade.Tests/ProductionTriggersFacadeServiceSpecs/WhenGettingTriggersWithNoCit.cs
```diff
@@ -0,0 +1,28 @@
...
+    using FluentAssertions;
+    using Linn.Common.Facade;
+    using Linn.Production.Domain.LinnApps.Measures;
+    using Linn.Production.Domain.LinnApps.Triggers;
+    using NSubstitute;
+    using NUnit.Framework;
+
+    public class WhenGettingTriggersWithNoCit : ContextBase
+    {
+
```
richardiphillips: again I think these should all be domain tests.

src/Service/Modules/ProductionTriggersModule.cs
```diff
@@ -0,0 +1,32 @@
...
+            this.Get("/production/reports/triggers", _ => this.GetProductionTriggers());
+        }
+
+        private object GetProductionTriggers()
+        {
+            var resource = this.Bind<ProductionTriggersRequestResource>();
+
+            var results =
+                this.productionTriggersFacadeService.GetProductionTriggerReport(resource.Jobref, resource.CitCode,
+                    resource.ReportType);
```
richardiphillips: does this module have tests?

PR #90, opened 2019-09-03
client for report

richardiphillips:


src/Service.Host/client/src/components/reports/AssemblyFailsMeasures.js
```diff
@@ -0,0 +1,45 @@
...
+import Grid from '@material-ui/core/Grid';
+import PropTypes from 'prop-types';
+import Box from '@material-ui/core/Box';
+import Page from '../../containers/Page';
+
+const AssemblyFailsMeasures = ({ reportData, loading, options }) => (
+    <Page>
+        <Grid container spacing={3} justify="center">
+            <Grid item xs={12}>
+                <Box paddingBottom={3}>
```
atcampbell: Could you put this padding on the grid so you could do without the Box?

richardiphillips: Can you pad a grid? I'm not sure. I'll probably remove this Box in the next iteration anyway.

PR #91, opened 2019-09-09
Trigger Report UI

andrewfraser73:
Initial version of production triggers report available as a drill down from OSR or directly from menu.

Featuring:
- reportType BRIEF/FULL now handled client side by component
- taking out invalid CITs from the drop down list (would have done this universally but not sure on other uses of CIT)
- a ContextMenu component that might be use in more places than here or may get scrapped if not enough functionality when users review app
- NotesPopover for story info that also may be useful in other places or might not survive into the final version
- ability to get production triggers without a jobref or a cit code and it will use sensible defaults (master PTL jobref or first production CIT).  This means when calling production triggers report from menu it populates with a default selection of options which you then change to suit your requirements.


src/Service.Host/client/src/components/reports/triggers/ProductionTriggers.js
```diff
@@ -0,0 +1,131 @@
...
+                                              )}`
+                                            : ''
+                                    }
+                                />
+                            </Grid>
+                        </Grid>
+                    ) : (
+                        ''
+                    )}
+                    {reportData ? (
```
atcampbell: Would this be clearer in the `reportData` conditional above?

src/Service.Host/client/src/components/reports/triggers/TriggersList.js
```diff
@@ -0,0 +1,87 @@
...
+                                    {m.story}
+                                </NotesPopover>
+                            ) : (
+                                ''
+                            )}
+                        </TableCell>
+                        <TableCell>
+                            <ContextMenu id={`triggers-menu${m.partNumber}`}>
+                                <span>
+                                    <a href={`facts?jobref=${jobref}&part-number=${m.partNumber}`}>
```
atcampbell: This could be a `Link` to keep the same styling

src/Facade/Services/ProductionTriggersFacadeService.cs
```diff
@@ -6,6 +6,7 @@
     using Linn.Production.Domain.LinnApps.Measures;
     using Linn.Production.Domain.LinnApps.Triggers;
     using Linn.Production.Persistence.LinnApps.Repositories;
+    using Remotion.Linq.Parsing.ExpressionVisitors.Transformation.PredefinedTransformations;
```
lewisrenfrew: looks suspect

PR #96, opened 2019-09-16
Works Orders route update

atcampbell:
Change works orders routes to use `/production/works-orders` 

Add auth to create and add routes

### DIFF
--- a/src/Facade/ResourceBuilders/WorksOrderDetailsResourceBuilder.cs
+++ b/src/Facade/ResourceBuilders/WorksOrderDetailsResourceBuilder.cs
@@ -24,7 +24,7 @@
 
         public string GetLocation(WorksOrderDetails worksOrderDetails)
         {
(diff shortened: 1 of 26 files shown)


src/Facade/ResourceBuilders/WorksOrderDetailsResourceBuilder.cs
```diff
@@ -24,7 +24,7 @@ public WorksOrderDetailsResource Build(WorksOrderDetails worksOrderDetails)
 
         public string GetLocation(WorksOrderDetails worksOrderDetails)
         {
-            return $"/production/maintenance/works-orders/details/{worksOrderDetails.PartNumber}";
+            return $"/production/works-orders/details/{worksOrderDetails.PartNumber}";
```
richardiphillips: what is a works order detail?

atcampbell: Its a route to get the audit message, part description and work station for the part. Could this do with a clearer name?

PR #109, opened 2019-09-20
Ptl settings

richardiphillips:
ptl settings form with authorisation


src/Service.Host/client/src/components/Root.js
```diff
@@ -115,6 +117,7 @@ const Root = ({ store }) => (
                                         />
                                         <Route
                                             exact
+
                                             path="/production/works-orders/outstanding-works-orders-report/report"
```
atcampbell: Extra blank lines

richardiphillips: yuk, don't know where they came from 👍

PR #115, opened 2019-10-02
production back orders added into facts

andrewfraser73:
production back orders are needed for trigger facts and will also be needed in OSR production back order report

### DIFF
--- a/src/Facade/Services/ProductionTriggersFacadeService.cs
+++ b/src/Facade/Services/ProductionTriggersFacadeService.cs
@@ -14,7 +14,7 @@
 
     public class ProductionTriggersFacadeService : IProductionTriggersFacadeService
     {
-        private readonly Domain.LinnApps.Repositories.IQueryRepository<ProductionTrigger> repository;
+        private readonly IQueryRepository<ProductionTrigger> repository;
 
         private readonly IMasterRepository<PtlMaster> masterRepository;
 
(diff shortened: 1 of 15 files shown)


src/Facade/Services/ProductionTriggersFacadeService.cs
```diff
@@ -19,12 +22,18 @@ public class ProductionTriggersFacadeService : IProductionTriggersFacadeService
 
         private readonly IRepository<WorksOrder, int> worksOrderRepository;
 
-        public ProductionTriggersFacadeService(Domain.LinnApps.Repositories.IQueryRepository<ProductionTrigger> repository, IRepository<Cit, string> citRepository, IMasterRepository<PtlMaster> masterRepository, IRepository<WorksOrder, int> worksOrderRepository)
+        private readonly IQueryRepository<ProductionBackOrder> productionBackOrderRepository;
+
+        private readonly IRepository<AccountingCompany, string> accountingCompaniesRepository;
+
+        public ProductionTriggersFacadeService(Domain.LinnApps.Repositories.IQueryRepository<ProductionTrigger> repository, IRepository<Cit, string> citRepository, IMasterRepository<PtlMaster> masterRepository, IRepository<WorksOrder, int> worksOrderRepository, IQueryRepository<ProductionBackOrder> productionBackOrderRepository, IRepository<AccountingCompany, string> accountingCompaniesRepository)
```
lewisrenfrew: Domain.Linnapps.Repositories prefix might not be needed here?

src/Facade/Services/ProductionTriggersFacadeService.cs
```diff
@@ -77,6 +86,18 @@ public IResult<ProductionTriggerFacts> GetProductionTriggerFacts(string jobref,
                     w.PartNumber == partNumber & w.DateCancelled == null & w.Quantity > w.QuantityBuilt).ToList()
                 : new List<WorksOrder>();
 
+            if (trigger.ReqtForSalesOrdersBE > 0)
+            {
+                var linn = this.accountingCompaniesRepository.FindById("LINN");
+                facts.OutstandingSalesOrders = linn != null
+                    ? this.productionBackOrderRepository.FilterBy(o =>
+                        o.JobId == linn.LatestSosJobId && o.ArticleNumber == partNumber).ToList() : new List<ProductionBackOrder>();
```
richardiphillips: Should this be in the domain? Should at least be a facade test for it?

andrewfraser73: I may write a test for it but it is really just marshalling all the components needed for a report and doesn't feel too business logic-y.

PR #125, opened 2019-10-11
Why Oh Why UI

andrewfraser73:
This is phase 1 of the Why PTL facts page of the trigger run

It also contains some changes to the backend to bring in enhanced trigger level and fixed build info.

It still isn't the finished article but because factory folk are looking at it I want to get it out there so it will be more correct for them looking at it.

The final version will have tests and ptl where used analysis


src/Service.Host/client/src/components/reports/triggers/SalesOrderList.js
```diff
@@ -0,0 +1,49 @@
...
+                    <TableCell>Qty On Order</TableCell>
+                    <TableCell>Requested Delivery Date</TableCell>
+                </TableRow>
+            </TableHead>
+            <TableBody>
+                {salesOrders.map(o => (
+                    <TableRow>
+                        <TableCell>
+                            <a href={`../../../sales/orders/${o.orderNumber}`}>{o.orderNumber}</a>
+                        </TableCell>
```
lewisrenfrew: would something involving APP_ROOT work here as opposed to the relative path?

andrewfraser73: Yes I'm not 100% happy with these hyperlinks and would prefer to use Link but I've had issues making them work.

src/Service.Host/client/src/reportTypes.js
```diff
@@ -36,6 +36,12 @@ export const productionTriggersReport = new ItemType(
     '/production/reports/triggers'
 );
 
+export const productionTriggerFacts = new ItemType(
+    'productionTriggerFacts',
+    'PRODUCTION_TRIGGER_FACTS',
+    '/production/reports/triggers/facts'
```
lewisrenfrew: could maybe make this actionTypeRoot 'PRODUCTION_TRIGGER_FACTS_REPORT'? A bit pedantic maybe but sticking to convention can't hurt

PR #134, opened 2019-10-22
Overdue orders report back end

atcampbell:


src/Domain.LinnApps/ViewModels/OverdueOrderLine.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public int NoStockQuantity { get; set; }
+
+        public int? AllocVal { get; set; }
+
+        public int? RnSDVal { get; set; }
+
+        public int? RnSNdVal { get; set; }
+
+        public int? RnNsVal { get; set; }
```
andrewfraser73: You could rename some of these columns to be  verbose on the acronymns we know like "RequiredNowNoStockValue"

andrewfraser73: Rtm = RequiredThisMonth, Rnm = RequiredNextMonth

src/Service/Modules/Reports/OrdersReportsModule.cs
```diff
@@ -32,5 +34,27 @@ private object ManufacturingCommitDateReport()
...
+        private object OverdueOrdersReportOptions()
+        {
+            return this.Negotiate.WithModel(ApplicationSettings.Get()).WithView("Index");
+        }
+
+        private object OverdueOrdersReport()
+        {
+            var resource = this.Bind<OverdueOrdersReportRequestResource>();
+            var results = this.reportService.GetOverdueOrdersReport(
+                resource.JobId,
```
andrewfraser73: Do you have to specify the Job Id in the overdue orders request or can you just get it from the Accounting Company?  I don't specify the job in production back orders but just get the latest one from that accounting company

src/Domain.LinnApps/ViewModels/OverdueOrderLine.cs
atcampbell: What would the Nd and a single D stand for?

src/Service/Modules/Reports/OrdersReportsModule.cs
atcampbell: Ahh yeah I'll have a look at that

src/Domain.LinnApps/ViewModels/OverdueOrderLine.cs
atcampbell: like RnSDVal and RnSNdVal

andrewfraser73: ND = Not Despatchable
D = Dispatchable 

So RnSDVal is Required Now Stock Despatachable Value i.e. the Can Go value from sales stats

RnSNdVal is Required Now Stock Not Despatchable Value i.e. is the Could go value and there is another column with reasons which can be multiple including Supply In Full, Credit etc

does this help?

atcampbell: Thanks!

PR #135, opened 2019-10-22
Sernos renum pack refactor

atcampbell:


atcampbell:
> and remove the reference to Resources from the the project (.csproj) too...

I've just realised that removing this from the csproj breaks the SalesArticleProxy class - which is getting a sales article from products maint & deserializes the json response. Should this class be changed as well?

richardiphillips:
Why does removing it from domain affect the proxy class?
We're a bit hamstrung with the proxy project because the restlcient classes rely on knowing about the external resource being returned to them. We should arguably keep those resources separately (maybe in the proxy project). I'm up for that change if somebody's keen but not immediately worried about it.

atcampbell:
> Why does removing it from domain affect the proxy class?
> We're a bit hamstrung with the proxy project because the restlcient classes rely on knowing about the external resource being returned to them. We should arguably keep those resources separately (maybe in the proxy project). I'm up for that change if somebody's keen but not immediately worried about it.

Oh sorry, I just meant to proxy project - should have been clearer. I'll keep it in there and remove it from domain.

PR #144, opened 2019-10-29
AWS yaml memory increase

atcampbell:
I've increased memory to 250 as this is what we're using in products-maint - does this seem like a reasonable number?

I've also removed the optional cpu values which may have been incorrect.

### DIFF
--- a/aws/application.yml
+++ b/aws/application.yml
@@ -109,12 +109,11 @@ Resources:
       TaskRoleArn: !GetAtt productionRole.Arn
       ContainerDefinitions:
       - Name: production
-        Cpu: '1'
         Essential: 'true'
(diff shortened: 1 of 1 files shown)


aws/application.yml
```diff
@@ -165,12 +164,11 @@ Resources:
...
       ContainerDefinitions:
       - Name: productionMessaging
-        Cpu: '1'
         Essential: 'true'
         Image: !Sub docker.io/linn/production-messaging:${dockerTag}
         PortMappings:
         - ContainerPort: 5050
-        Memory: '150'
+        Memory: '250'
         LogConfiguration:
```
lewisrenfrew: does the messaging service need more memory too?

PR #146, opened 2019-10-30
Batch notes viewer

lewisrenfrew:
- I spoke to Stewart about this form and he said all he uses it for is to search works orders by part number to get a quick overview.

- I made a new component that behaves very similarly to the TypeAhead except it shows results as a table with headers as opposed to just a 'Name - Description' list. Clicking on a table row takes you through to that rows self-href in the usual way. 

- I'll make this a shared component if everyone approves although anywhere that this component could be used is already catered for by the ReportOptions -> ReportTable method. I thought this interactive search table was a bit more usable for his purposes though.

- For some searches, e.g. searching 'PCAS' there are tens of thousands of results, So I've just returned a bad request in this case (greater than 500 results) to stop the client getting unresponsive.


src/Service.Host/client/src/components/worksOrders/TypeAheadTable.js
```diff
@@ -0,0 +1,131 @@
...
+    title,
+    loading,
+    clearSearch,
+    history,
+    placeholder
+}) {
+    const [searchTerm, setSearchTerm] = useState('');
+    useSearch(fetchItems, searchTerm, clearSearch);
+
+    const handleSearchTermChange = (...args) => {
```
atcampbell: I've done this as well, but I think Richard somewhere did `(_propertName, newValue)` to show what the parameter is (as well as the _ denoting an unused parameter) - I think that's clearer than doing `(...args)`

src/Service.Host/client/src/components/worksOrders/WorksOrder.js
```diff
@@ -124,7 +124,7 @@ function WorksOrder({
 
     const handleBackClick = () => {
         setEditStatus('view');
-        history.push('/production/works-orders');
+        history.push('/production/works-orders-batch-notes');
```
atcampbell: I'm not sure if this would always be correct either as it will give the wrong back link from the works order utility. I'm not sure what is the best to do without a context sensitive back -  but maybe just whatever one is the most used?

src/Service.Host/client/src/itemTypes.js
```diff
@@ -92,6 +92,12 @@ export const worksOrder = new ItemType('worksOrder', 'WORKS_ORDER', '/production
 
 export const worksOrders = new ItemType('worksOrders', 'WORKS_ORDERS', '/production/works-orders');
 
+export const worksOrdersBathcNotes = new ItemType(
```
atcampbell: Typo in the item type name

src/Service/Modules/WorksOrdersModule.cs
```diff
@@ -31,6 +33,7 @@ public WorksOrdersModule(IOutstandingWorksOrdersReportFacade outstandingWorksOrd
 
             this.Get("/production/works-orders/outstanding-works-orders-report", _ => this.GetOutstandingWorksOrdersReport());
             this.Get("/production/works-orders/outstanding-works-orders-report/export", _ => this.GetOutstandingWorksOrdersReportExport());
+            this.Get("/production/works-orders-for-part", _ => this.GetWorksOrdersForPart());
```
atcampbell: Would `/produciton/works-orders/works-orders-for-part` be more fitting with the existing routes?

src/Service.Host/client/src/components/worksOrders/WorksOrder.js
atcampbell: Or maybe it could go to `/production/works-orders` & the works order search page could have a button to go to the batch viewer search?

PR #150, opened 2019-10-31
Truefacts

andrewfraser73:
clean version of my why why oh changes with links and where used assemblies


src/Domain.LinnApps/Triggers/ProductionTriggerAssembly.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public int? RemainingBuild { get; set; }
+
+        public int? ReqtForInternalAndTriggerLevelBT { get; set; }
+
+        public int? ReqtForPriorityBuildBE { get; set; }
+
+        public bool HasReqt()
+        {
+            //  (PW.NETT_SALES_ORDERS != 0 OR PW.QTY_BEING_BUILT != 0 OR PW.BT != 0)
```
richardiphillips: could probably take these comments out.

PR #152, opened 2019-10-31
batch notes viewer tweaks

lewisrenfrew:
- only show board part works orders

- truncate results when there's too many as opposed to returning a bad request


src/Domain.LinnApps/Part.cs
```diff
@@ -35,5 +35,10 @@ public bool IsPhantomPart()
         {
             return this.BomType == "P";
         }
+
+        public bool IsBoardPart()
+        {
+            return this.PartNumber.StartsWith("PCAS") || this.PartNumber.StartsWith("PCSM");
```
richardiphillips: Is there a test for this?

lewisrenfrew: there will be very soon

Issue #44, opened 2019-07-23
WO Labels


andrewfraser73:
Production Planning
Labels
WO Labels
WO_LABEL_UT
Could still be needed - check

PR #155, opened 2019-11-04
Wo labels ut

lewisrenfrew:
- satisfies #44 

- I accidentally pushed the domain and persistence layers parts straight to master so the files aren't in this changeset. You can see them here if you want to double check: https://github.com/linn/production/commit/3769ec05f31e92c144dfb69d69ae64279a90a1e7


src/Service.Host/client/src/selectors/worksOrderLabelSelectors.js
```diff
@@ -1,4 +1,4 @@
 import { ItemSelectors } from '@linn-it/linn-form-components-library';
```
richardiphillips: is this file deliberately renamed?

src/Service/Modules/WorksOrdersModule.cs
```diff
@@ -18,12 +21,22 @@ public sealed class WorksOrdersModule : NancyModule
...
+            IFacadeService<WorksOrderLabel, WorksOrderLabelKey, WorksOrderLabelResource, WorksOrderLabelResource> labelService, 
+            IWorksOrdersService worksOrdersService)
         {
             this.worksOrdersService = worksOrdersService;
+            this.labelService = labelService;
             this.outstandingWorksOrdersReportFacade = outstandingWorksOrdersReportFacade;
 
             this.Get("/production/works-orders", _ => this.GetWorksOrders());
+            this.Get("/production/works-orders/labels", _ => this.GetWorksOrderLabelsForPart());
+            this.Get("/production/works-orders/labels/{seq}/{part*}", parameters => this.GetWorksOrderLabel(parameters.part, parameters.seq));
```
richardiphillips: bit dodgy this uri but I can see why you did it.

PR #156, opened 2019-11-06
Manufacturing routes authorisation checks for update & add + tests + fixes

iaincrawf:


src/Facade/ResourceBuilders/ManufacturingRoutesResourceBuilder.cs
```diff
@@ -1,26 +1,37 @@
...
+        private readonly ManufacturingRouteResourceBuilder manufacturingRouteResourceBuilder;
 
-        public IEnumerable<ManufacturingRouteResource> Build(IEnumerable<ManufacturingRoute> manufacturingRoutes)
+        public ManufacturingRoutesResourceBuilder(IAuthorisationService authorisationService)
         {
+            this.AuthorisationService = authorisationService;
+            this.manufacturingRouteResourceBuilder = new ManufacturingRouteResourceBuilder(this.AuthorisationService);
+        }
+
+        protected IAuthorisationService AuthorisationService { get; }
```
atcampbell: I think you can remove this from this class?

src/Facade/Services/ManufacturingOperationsService.cs
```diff
@@ -5,23 +5,27 @@
...
         {
+            this.databaseService = databaseService;
         }
 
         protected override ManufacturingOperation CreateFromResource(ManufacturingOperationResource resource)
         {
             return new ManufacturingOperation(
                 resource.RouteCode,
-                resource.ManufacturingId,
+                this.databaseService.GetIdSequence("AN_ARBITRARY_SEQUENCE"),
```
atcampbell: Is this bit unfinished?

src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
```diff
@@ -34,9 +35,10 @@ function ManufacturingRoute({
     const creating = () => editStatus === 'create';
     const editing = () => editStatus === 'edit';
     const viewing = () => editStatus === 'view';
+    const allowedToEdit = item ? utilities.getHref(item, 'edit') !== null : false;
```
atcampbell: I find this statement a wee bit confusing, could it be split up? This might also be better as a selector and passed in from the container.

src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
```diff
@@ -34,9 +35,10 @@ function ManufacturingRoute({
...
 
     useEffect(() => {
-        if (item) {
+        if (item && !(editStatus === 'create')) {
```
atcampbell: May as well use the creating() method above

src/Service/Modules/ManufacturingRoutesModule.cs
```diff
@@ -42,24 +57,56 @@ private object GetById(string routeCode)
...
+                        .WithView("Index");
+                }
+                catch (Exception e)
+                {
+                    return this.Negotiate.WithModel(new BadRequestResult<Error>(e.Message));
+                }
+            }
+            else
+            {
+                return HttpStatusCode.Forbidden;
```
atcampbell: You can return an UnauthorisedResult here - will be something like `return this.Negotiate.WithModel(new UnauthorisedResult<SaHoldStory>("You are not authorised to create hold stories."));`

src/Service/Modules/ManufacturingRoutesModule.cs
```diff
@@ -42,24 +57,56 @@ private object GetById(string routeCode)
...
```
atcampbell: UnauthorisedResult

tests/Integration/Service.Tests/ManufacturingRoutesModuleSpecs/WhenGettingById.cs
```diff
@@ -17,8 +17,13 @@ public void SetUp()
         {
             var route = new ManufacturingRoute("TESTCODE", "desc", "a note");
             route.Operations = new List<ManufacturingOperation>();
-            this.ManufacturingRouteService.GetById("TESTCODE")
-                .Returns(new SuccessResult<ManufacturingRoute>(route));
+
+
+            //this.AuthorisationService.HasPermissionFor(AuthorisedAction.ManufacturingRouteUpdate, Arg.Any<List<string>>())
```
atcampbell: comments

tests/Unit/Facade.Tests/ManufacturingOperationsServiceSpecs/WhenAddingManufacturingOperation.cs
```diff
@@ -59,7 +58,7 @@ public void ShouldReturnCreated()
             this.result.Should().BeOfType<CreatedResult<ManufacturingOperation>>();
             var dataResult = ((CreatedResult<ManufacturingOperation>)this.result).Data;
             dataResult.RouteCode.Should().Be(this.manufacturingOperation.RouteCode);
-            dataResult.ManufacturingId.Should().Be(this.manufacturingOperation.ManufacturingId);
+            //dataResult.ManufacturingId.Should().Be(this.manufacturingOperation.ManufacturingId);
```
atcampbell: comments

src/Facade/ResourceBuilders/ManufacturingRoutesResourceBuilder.cs
iaincrawf: It's used to pass through to route builder in the constructor, but should I just pass through the local authorisationService to that builder? Stylecop made me put it below the constructor

src/Facade/Services/ManufacturingOperationsService.cs
iaincrawf: Nah, the procedure is just actually called "AN_ARBITRARY_SEQUENCE" hahaha
image

src/Facade/ResourceBuilders/ManufacturingRoutesResourceBuilder.cs
richardiphillips: you can just pass authorisationService parameter straight into the builder. You don't need the variable at all.

richardiphillips: In fact maybe you could just pass in the builder interface itself

PR #157, opened 2019-11-07
Wwdbackend

andrewfraser73:
Backend for What Will Decrement Report from trigger drilldown.

Also contains helper functions for calling oracle procedures and functions.

### DIFF
--- a/src/Facade/Services/WwdResultFacadeService.cs
+++ b/src/Facade/Services/WwdResultFacadeService.cs
@@ -47,12 +47,12 @@
             var result = new WwdResult
             {
                 PartNumber = partNumber,
-                Qty = (int) qty,
+                Qty = qty.Value,
                 WorkStationCode = triggerLevel.WsName,
                 PtlJobref = ptlJobref
             };
 
-            result.WwdJobId = this.wwdTrigFunction.WwdTriggerRun(partNumber, (int) qty);
+            result.WwdJobId = this.wwdTrigFunction.WwdTriggerRun(partNumber, qty.Value);
 
             if (result.WwdJobId == 0)
             {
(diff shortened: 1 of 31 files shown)


src/Facade/Services/WwdResultFacadeService.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            if (string.IsNullOrEmpty(triggerLevel.WsName))
+            {
+                return new NotFoundResult<WwdResult>("No work station found");
+            }
+
+            var result = new WwdResult
+            {
+                PartNumber = partNumber,
+                Qty = (int) qty,
```
richardiphillips: qty.Value - don't need the cast

src/Facade/Services/WwdResultFacadeService.cs
```diff
@@ -0,0 +1,68 @@
...
+                WorkStationCode = triggerLevel.WsName,
+                PtlJobref = ptlJobref
+            };
+
+            result.WwdJobId = this.wwdTrigFunction.WwdTriggerRun(partNumber, (int) qty);
```
richardiphillips: .Value as before

tests/Unit/Facade.Tests/WwdResultFacadeServieSpecs/WhenGeneratingWwdResult.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace Linn.Production.Facade.Tests.WwdResultFacadeServieSpecs
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
```
richardiphillips: what is this?

Issue #48, opened 2019-07-24
Manufacturing Routes Utility


andrewfraser73:
Production Maintenance
Labour, Skills and Resources Planning
Manufacturing Routes Utility
MFG_ROUTES_UT

PR #159, opened 2019-11-11
Manufacturing routes updates

iaincrawf:
Updated backend for deletion of operations, updated front end for delete and add and errors

#48


src/Facade/Services/IManufacturingOperationsService.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Linn.Production.Facade.Services
+{
+    using Linn.Common.Facade;
+    using Linn.Production.Domain.LinnApps;
+    using Linn.Production.Resources;
+
+    public interface IManufacturingOperationsService : IFacadeService<ManufacturingOperation, int,
+        ManufacturingOperationResource, ManufacturingOperationResource>
+    {
+        void RemoveOperation(ManufacturingOperation entity);
```
atcampbell: I think this should return an `IResult`?

src/Service.Host/client/src/containers/manufacturingRoutes/ManufacturingRoutes.js
```diff
@@ -11,9 +11,7 @@ const mapStateToProps = state => ({
     itemError: getItemError(state, itemTypes.manufacturingRoutes.item)
 });
 
-const initialise = () => () => {
-    // dispatch(manufacturingRoutesActions.fetch());
-};
+const initialise = () => () => {};
```
atcampbell: If this is an empty function it and `initialiseOnMount` can be removed

richardiphillips: agree with AC

PR #164, opened 2019-11-15
Assembly fails details report and export.

### LINKED ISSUE #21: Assembly Fails By Product
Assembly Quality System
Assembly Fails By Product
ASSEMBLY_FAILS_BY_PRODUCT
Replace this and the next two with a more general csv output with limited options (ASSEMBLY_FAILS_BY_BOARD, ASSEMBLY_FAILS_BY_FAILURE)

richardiphillips:
closes #21


src/Service.Host/client/src/components/reports/AssemblyFailsDetailsOptions.js
```diff
@@ -0,0 +1,69 @@
...
+                    </Typography>
+                </Grid>
+                <Grid item xs={3}>
+                    <DatePicker label="From Date" value={fromDate} onChange={setFromDate} />
+                </Grid>
+                <Grid item xs={3}>
+                    <DatePicker
+                        label="To Date"
+                        value={toDate}
+                        minDate={fromDate}
```
lewisrenfrew: DatePicker might complain about prop types if you give it Date objects for value, minDate or maxDate. It expects strings if I  recall correctly (we could change it to accept both and toString() internally if needed). The actual mui component is fine with date objects so it doesn't really matter... But in the interest of a warnings free life you could just toString() from here for now.

PR #168, opened 2019-11-18
What Will Decrement UI

andrewfraser73:
What will decrement UI with coloured icons for no stock and reserved stock
had to make some changes to triggers and back end to make this work
Only 16 files


src/Facade/Services/WwdResultFacadeService.cs
```diff
@@ -49,7 +50,8 @@ public IResult<WwdResult> GenerateWwdResultForTrigger(string partNumber, int? qt
                 PartNumber = partNumber,
                 Qty = qty.Value,
                 WorkStationCode = triggerLevel.WsName,
-                PtlJobref = ptlJobref
+                PtlJobref = ptlJobref,
+                WwdRunTime = DateTime.Now
```
richardiphillips: generally stick to UtcNow for consistency

src/Service.Host/client/src/components/reports/wwd/WwdTriggerReport.js
```diff
@@ -0,0 +1,94 @@
...
+                )}
+                {itemError ? (
+                    <ErrorCard errorMessage={itemError.details?.message} />
+                ) : (
+                    <Grid item xs={12}>
+                        {reportData ? (
+                            <Fragment>
+                                <Title text="What Will Decrement From A Workstation" />
+                                <div className={classes.padbottom}>
+                                    <span>
```
atcampbell: Could this be better as a MUI Typography rather than a span? (I think that might have more consistent styling?)

src/Facade/Services/WwdResultFacadeService.cs
richardiphillips: Should be a test change for this too somewhere?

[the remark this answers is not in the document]
andrewfraser73: Do you mean PtlJobref and WwdRunTime - it is just using abbreviations with Wwd is WhatWillDecrement and Ptl is Production Trigger Level

Or is it something else that is unclear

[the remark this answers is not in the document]
atcampbell: Yeah just thinking about avoiding the abbreviations, these one's aren't too bad though

PR #178, opened 2019-11-20
Assembly Fail Fault Code create route fix

atcampbell:
I've added a check for a fault code /create route - I was catching all possible ids as many have `/` characters which weren't properly caught by nancy. 

I went half way down the route of using a request parameter `?faultCode="bleh"` but I stopped as I wasn't sure how this would work in a conditional with react router.

If the request param seems cleaner I can implement that instead.


src/Service/Modules/AssemblyFailsModule.cs
```diff
@@ -89,6 +89,11 @@ private object GetFaultCodes()
 
         private object GetFaultCode(string id)
         {
+            if (id.ToLower().Equals("create"))
```
richardiphillips: this doesn't feel good

atcampbell: Yeah after looking at Iains pr I think it's better to check for the route specifically and have a get app method?

PR #179, opened 2019-11-20
added styles for printing to fit large table on a4 paper

iaincrawf:
Any thoughts on where to put the css file or on improving this? Feel free to give me a shout about this in general. Will also need to up version no.


iaincrawf:
@richardiphillips I've actually just tested it and I could put it into the page component just straight in the shared components library. Do you think it's better to put it in there and have everywhere covered rather than a case by case basis? My only concerns would be that it'll zoom out every table slightly (seems fine on all that I've tested so far, probably something we want actually) and that I've locked it in portrait orientation (seems to mess everything up if I take that off and let it go landscape though)

PR #182, opened 2019-11-22
update prettier and obey its new rules.

richardiphillips:


richardiphillips:
No functional changes. Just linting.

PR #184, opened 2019-11-26
Label type utility (still need to add front end tests, forgot about that)

iaincrawf:
is production/resources where this uri should be or somewhere else like production/maintenance?


src/Service.Host/client/src/components/labelTypes/LabelTypes.js
```diff
@@ -0,0 +1,144 @@
...
+            setRowsToDisplay([]);
+        } else {
+            setRowsToDisplay(
+                rows
+                    .sort(compare(pageOptions.orderBy, pageOptions.orderAscending))
+                    .slice(
+                        pageOptions.currentPage * pageOptions.rowsPerPage,
+                        pageOptions.currentPage * pageOptions.rowsPerPage + pageOptions.rowsPerPage
+                    )
+            );
```
lewisrenfrew: Doesn't each rowToDisplay need an id field? Think paginated table will complain without them. You can just do 

`.map(row => ({...row, id: row.labelTypeCode}))`

as long as labelTypeCode is always non null and unique

src/Service.Host/client/src/containers/labelTypes/CreateLabelType.js
```diff
@@ -0,0 +1,28 @@
...
+    item: {},
+    editStatus: 'create',
+    itemError: getItemError(state, itemTypes.labelType.item),
+    loading: labelTypeSelectors.getLoading(state),
+    snackbarVisible: labelTypeSelectors.getSnackbarVisible(state)
+});
+
+const initialise = () => dispatch => {
+    dispatch(labelTypeActions.clearErrorsForItem());
+    dispatch(labelTypeActions.setEditStatus('create'));
```
lewisrenfrew: probably unecessary? editStatus is already create

src/Service.Host/client/src/components/labelTypes/LabelTypes.js
iaincrawf: yep, I forgot to fix that, will add it in!

PR #189, opened 2019-11-28
Builds by dept report fixes

lewisrenfrew:


src/Domain.LinnApps/Reports/BuildsDetailReportService.cs
```diff
@@ -49,24 +49,25 @@ public class BuildsDetailReportService : IBuildsDetailReportService
 
             foreach (var partGroup in partGroups)
             {
-                var partTotal = new decimal();
+                var partTotal = new decimal(0);
```
richardiphillips: I'd never thought of newing up a decimal. I'd do decimal partTotal = 0 or var partTotal = 0m if it were me

PR #192, opened 2019-12-03
fix content types that are being returned as application/json

richardiphillips:


src/Service/ResponseProcessors/AssemblyFailResponseProcessor.cs
```diff
@@ -7,7 +7,7 @@
     public class AssemblyFailResponseProcessor : JsonResponseProcessor<AssemblyFail>
     {
         public AssemblyFailResponseProcessor(IResourceBuilder<AssemblyFail> resourceBuilder)
-            : base(resourceBuilder)
+            : base(resourceBuilder, "assembly-fails-fault_codes", 1)
```
lewisrenfrew: possible rogue _ ?

PR #193, opened 2019-12-04
Label reissue privileges

richardiphillips:
Proposed HATEOAS implementation of privileges where no item or collection has been requested.

Will remove incubated js factories before merge.


src/Service/Modules/LabelsModule.cs
```diff
@@ -91,7 +103,10 @@ private object ReprintAllLabels()
 
         private object GetApp()
         {
-            return this.Negotiate.WithModel(ApplicationSettings.Get()).WithView("Index");
+            return this.Negotiate
+                .WithModel(new SuccessResult<ResponseModel<LabelReprint>>(new ResponseModel<LabelReprint>(new LabelReprint(), new List<string>())))
```
iaincrawf: Could you talk me/us through this at some point? I'm unsure if this is to solve the problem I had before or something different, but if it's the same sort of 'get app without any specific items' scenario I can't figure out where in the code you're calling and returning the privileges from - thought it'd have been here in GetApp instead of new `List<string>()`

richardiphillips: Good spot. i noticed this last night. It's what I get for publishing a PR ahead of actually making it work!

PR #195, opened 2019-12-05
Trigger levels utility (v1)

iaincrawf:
Need to go and speak to stephen once this is up for any changes


src/Domain.LinnApps/ProductionTriggerLevel.cs
```diff
@@ -2,6 +2,9 @@
 {
     public class ProductionTriggerLevel
     {
+        public ProductionTriggerLevel()
+        {
```
richardiphillips: don't need this public constructor

PR #199, opened 2019-12-10
Works Order serial numbers view

atcampbell:
- Add table to view serial numbers associated with a works order
- Issue Serial Number takes a query string to search sernos on mount
- Link to Issue Sernos from works order serial numbers


src/Service.Host/client/src/components/worksOrders/WorksOrder.js
```diff
@@ -233,7 +245,7 @@ function WorksOrder({
                     )}
                 </Grid>
                 {itemErrors &&
-                    itemErrors.map(itemError => (
+                    itemErrors.map(() => (
                         <Grid item xs={12}>
```
lewisrenfrew: no map key here... not sure what you'd use as the map key though. Array index might be your only option (and not necessarily a bad one since itemErrors won't be reordered.)

PR #200, opened 2019-12-10
updated trigger levels ut

iaincrawf:
Need to ask lewis about typeahead id console warnign when running test + see if he can get/focus on a dropdown element in another test


src/Service.Host/client/src/components/__tests__/ProductionTriggersTests/productionTriggerLevels.test.js
```diff
@@ -63,9 +63,24 @@ describe('When viewing', () => {
...
     });
+
+    // test('should display cit dropdown', () => {
+    //     const { getByDisplayValue } = render(<ProductionTriggerLevels {...defaultProps} />);
+    //     // const item = queryByRole('select');
+    //     // expect(item).toBeInTheDocument();
+    //     expect(getByDisplayValue('Cit')).toBeInTheDocument();
+    // });
+    //cant get this to work no matter what I try to get the dropdown
+    //TODO - get help on this or get it working
```
lewisrenfrew: It can be really annoying selecting stuff using this library.... You could do a test Id or something if you really wanted but in this case I wouldn't worry too much about making sure the dropdown  is rendered. Unless its conditionally rendered based on some logic that differs tothat of the rest of the components I think you can quite safely assume it renders along with everything else.

src/Facade/Services/ProductionTriggerLevelService.cs
```diff
@@ -57,7 +58,17 @@ protected override void UpdateFromResource(ProductionTriggerLevel entity, Produc
...
+            if (searchTerms.Length != 4)
+            {
+                return w => w.PartNumber.ToUpper().Contains(searchTerm.ToUpper());
+            }
+
+            return w =>
+                (string.IsNullOrWhiteSpace(searchTerms[0]) || w.PartNumber.ToUpper().Contains(searchTerms[0].ToUpper()))
+                && (string.IsNullOrWhiteSpace(searchTerms[1]) || w.CitCode == searchTerms[1])
+                && (searchTerms[2] == "null" || w.OverrideTriggerLevel > int.Parse(searchTerms[2]))
+                && (searchTerms[3] == "null" || w.VariableTriggerLevel > int.Parse(searchTerms[3]));
```
lewisrenfrew: Not 100% sure what's going on here

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
```diff
@@ -86,10 +90,11 @@ function TriggerLevel({
     };
 
     const handleBackClick = () => {
-        history.push('/production/resources/manufacturing-routes/');
+        history.push('/production/maintenance/production-trigger-levels/');
     };
 
     const handleResourceFieldChange = (propertyName, newValue) => {
+        console.log(`tryna set ${newValue} in ${propertyName}`);
         setTriggerLevel({ ...triggerLevel, [propertyName]: newValue });
```
lewisrenfrew: left a console log in

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
```diff
@@ -147,20 +152,23 @@ function TriggerLevel({
...
-                                                        ? 'This field is required'
-                                                        : ''
-                                                }
-                                                required
-                                                fullWidth
                                                 label="Part Number"
-                                                allowNoValue={false}
+                                                modal
+                                                items={partsSearchResults}
+                                                value={triggerLevel.partNumber}
```
lewisrenfrew: your console error thing might be because each item in items needs and id field. You can just map 

`items.map(i => ({...i, id: i.someUniqueField})` 

to achieve this.

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevels.js
```diff
@@ -92,16 +117,59 @@ const ViewProductionTriggerLevels = ({ loading, itemError, history, items, fetch
...
+                        allowNoValue
+                    />
+                </Grid>
+                <Grid item xs={1} />
+                <Grid item xs={1}>
+                    <SearchInputField
+                        label="Override Trigger >"
+                        fullWidth
+                        // placeholder="search.."
+                        onChange={handleOverrideSearchChange}
```
lewisrenfrew: are the placeholders commented out for a reason?

src/Service.Host/client/src/containers/productionTriggerLevels/ProductionTriggerLevel.js
```diff
@@ -19,7 +19,7 @@ const mapStateToProps = (state, { match }) => ({
     editStatus: productionTriggerLevelSelectors.getEditStatus(state),
     loading: productionTriggerLevelSelectors.getLoading(state),
     snackbarVisible: productionTriggerLevelSelectors.getSnackbarVisible(state),
-    parts: [],
+    partsSearchResults: [],
```
lewisrenfrew: is this supposed to be an empty array? without thinking about it too much i would have guessed this would come in from redux state.

src/Service.Host/client/src/containers/productionTriggerLevels/CreateProductionTriggerLevel.js
```diff
@@ -15,23 +15,24 @@ import workStationActions from '../../actions/workStationActions';
...
     manufacturingRoutes: manufacturingRoutesSelectors.getItems(state),
     cits: citsSelectors.getItems(state),
     employees: employeesSelectors.getItems(state),
     itemErrors: getItemError(state, itemTypes.productionTriggerLevel.item),
-    workStations: workStationSelectors.getItems(state)
+    workStations: workStationSelectors.getItems(state),
+    partsSearchResults: partsSelectors
+        .getSearchItems(state)
+        .map(s => ({ ...s, id: s.partNumber, name: s.partNumber })),
+    partsSearchLoading: partsSelectors.getSearchLoading(state)
```
lewisrenfrew: looks like you are providing an id here... Might be something else causing the error. I'll have a look and see if I can spot it.

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
```diff
@@ -388,36 +396,40 @@ TriggerLevel.propTypes = {
...
-    },
+    item: {},
     snackbarVisible: false,
     addItem: null,
     updateItem: null,
     loading: null,
     itemErrors: null,
     itemId: null,
-    parts: [{ partNumber: '', description: '' }]
+    partsSearchResults: [{ partNumber: '', description: '' }],
```
lewisrenfrew: Not really important but just for cleanliness - do you need to be this explicit with the default prop here? would null or an empty array suffice?

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
```diff
@@ -388,36 +396,40 @@ TriggerLevel.propTypes = {
...
+    workStations: [],
+    cits: [],
+    employees: [],
+    searchParts: () => {},
```
lewisrenfrew: again, seems odd to pass this as a default prop. I'd either make this required and ensure these functions are always supplied (they're just functions obviously so not waiting on anything arriving) or make it null and only render stuff that needs them (like the Typeahead) if you know they're going to be there. In the current form it implies default acceptable behaviour is just for the searches not to work even if they render which seems odd to me.

src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
iaincrawf: I basically only allow changing and looking up part numbers when you're creating, so I didn't want the normal triggerLevel container to have to be worried about part numbers stuff, think I was being that explicit to avoid console errors.. I can definitely tidy some of this up as I'm passing [ ] in the container and the above in default props and don't need both

lewisrenfrew: I think if it was me I would just make parts default to null and then only render stuff that complains when parts is null if parts is not null (if that makes sense). No harm in doing it the way you've done it either though.

iaincrawf: Ah - I had done it to fix test warnings but really I'd forgotten to pass in those for the create test, oops. Added them in and set the defaults to null

PR #207, opened 2019-12-12
Build plan utility service

atcampbell:


src/Facade/ResourceBuilders/BuildPlanDetailResourceBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        object IResourceBuilder<ResponseModel<BuildPlanDetail>>.Build(ResponseModel<BuildPlanDetail> model) =>
+            this.Build(model);
+
+        private IEnumerable<LinkResource> BuildLinks(ResponseModel<BuildPlanDetail> model)
+        {
+            yield return new LinkResource { Rel = "self", Href = this.GetLocation(model) };
+
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.BuildPlanAdd, model.Privileges))
```
iaincrawf: Feels like either the permission checks for add and update should be separate since there's a different AuthorisedAction for add & update or we don't need the 2 different AuthorisedActions

iaincrawf: Actually this one should be the buildPlanDetailAdd too shouldn't it?

src/Facade/ResourceBuilders/BuildPlanResourceBuilder.cs
```diff
@@ -19,20 +31,28 @@ public BuildPlanResource Build(BuildPlan buildPlan)
...
-        object IResourceBuilder<BuildPlan>.Build(BuildPlan buildPlan) => this.Build(buildPlan);
+        object IResourceBuilder<ResponseModel<BuildPlan>>.Build(ResponseModel<BuildPlan> buildPlan) => this.Build(buildPlan);
 
-        private IEnumerable<LinkResource> BuildLinks(BuildPlan buildPlan)
+        private IEnumerable<LinkResource> BuildLinks(ResponseModel<BuildPlan> model)
         {
-            yield return new LinkResource { Rel = "self", Href = this.GetLocation(buildPlan) };
+            yield return new LinkResource { Rel = "self", Href = this.GetLocation(model) };
+
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.BuildPlanAdd, model.Privileges))
```
iaincrawf: same with maybe removing the unused authorised action or making separate checks

src/Facade/ResourceBuilders/BuildPlanRuleResourceBuilder.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        object IResourceBuilder<ResponseModel<BuildPlanRule>>.Build(ResponseModel<BuildPlanRule> buildPlanRule) =>
+            this.Build(buildPlanRule);
+
+        private IEnumerable<LinkResource> BuildLinks(ResponseModel<BuildPlanRule> model)
+        {
+            yield return new LinkResource { Rel = "self", Href = this.GetLocation(model) };
+
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.BuildPlanAdd, model.Privileges))
```
iaincrawf: just wanted to check this is the right check also

src/Facade/Services/BuildPlanDetailService.cs
```diff
@@ -0,0 +1,49 @@
...
+                                      {
+                                          BuildPlanName = resource.BuildPlanName,
+                                          FromLinnWeekNumber = resource.FromLinnWeekNumber,
+                                          PartNumber = resource.PartNumber,
+                                          Quantity = resource.Quantity,
+                                          RuleCode = resource.RuleCode,
+                                          ToLinnWeekNumber = resource.ToLinnWeekNumber
+                                      };
+
+            buildPlanDetail.Validate();
```
iaincrawf: What does .Validate() do and is it built in? Just for me if I should be using it!

src/Facade/ResourceBuilders/BuildPlanDetailResourceBuilder.cs
atcampbell: Oooops, good catch - lots of lazy copy pasting I think 😱

src/Facade/Services/BuildPlanDetailService.cs
atcampbell: It's a method I've added to the BuildPlanDetail class that throws an exception if its invalid. I put that in instead of creating and updating them through a factory.

PR #214, opened 2019-12-20
Label printer partial back end

iaincrawf:
enough to build the front end but haven't done the actual printing bit yet


src/Facade/ResourceBuilders/IdAndNameResourceBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                Id = idAndName.Id,
+                Name = idAndName.Name,
+                Links = this.BuildLinks(idAndName).ToArray()
+            };
+        }
+
+        public string GetLocation(IdAndName idAndName)
+        {
+            return string.Empty;
```
lewisrenfrew: Is there a reason location should be an empty string? Could be misleading if not. Could throw a NotImplementedException to make it clearer.

src/Facade/ResourceBuilders/LabelPrintResourceBuilder.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    using Linn.Common.Authorisation;
+    using Linn.Common.Facade;
+    using Linn.Common.Resources;
+    using Linn.Production.Domain.LinnApps;
+    using Linn.Production.Resources;
+
+    public class LabelPrintResourceBuilder : IResourceBuilder<LabelPrint>
+    {
+        public LabelPrintResourceBuilder()
```
lewisrenfrew: Is there something making the empty constructor mandatory here?

src/Facade/Services/LabelPrintService.cs
```diff
@@ -0,0 +1,79 @@
+﻿namespace Linn.Production.Facade.Services
+
```
lewisrenfrew: blank

tests/Integration/Service.Tests/LabelPrintModuleSpecs/ContextBase.cs
```diff
@@ -0,0 +1,61 @@
...
+    using NSubstitute;
+
+    using NUnit.Framework;
+
+    public class ContextBase : NancyContextBase
+    {
+        protected ILabelPrintService LabelPrintService { get; private set; }
+        
+    [SetUp]
+        public void EstablishContext()
```
lewisrenfrew: weird indenation

src/Facade/ResourceBuilders/IdAndNameResourceBuilder.cs
iaincrawf: Nope good point, thinking about it I probably don't actually need to return any links with empty Hrefs either, think I'd be better just returning an empty list of Link Resources

Issue #204, opened 2019-12-12
Show works order comments


richardiphillips:
The 'old' works order viewer in linnapps shows an empty batch comments field but there is apparently no screen to add these comments any more.

Stewart MacAulay would like to be able to add these comments or any other kind of comment to the works order that would be displayed on the works order screens.

Issue #203, opened 2019-12-12
Batch viewer - back button


richardiphillips:
Hitting back goes right back to the menu. Would be nice if it went back to the search results.

Issue #201, opened 2019-12-12
batch viewer - search problems


richardiphillips:
searching for e.g. PCSM 973 brings back wrong results. Seems like > 1 searches are sent to server and the most general takes longest and overwrites the correct one. 

Probably the case elsewhere too

PR #219, opened 2020-01-07
Batch notes viewer fixes

lewisrenfrew:
Addressing:

#204 - Added the database field 'QUALITY_ISSUES' as a Batch Notes field on the WorksOrder domain object and display/allow editing on the form. Not 100% sure this is what Stewart wanted as the old Linnapps page was a bit dubious. Will see what he thinks.

#203 - This one was a bit of a problem because the Works Order Utility can be used from a few contexts, meaning hardcoding the back button to go back to Stewart's search results might annoy other users who don't necessarily want to go back to the same place as he does. I tried to make a contextual back button using the connected-react-router package to synchronise a 'history' redux store with browser history and keep track of a users path through the app. In this way we can use back buttons to direct the user back to the last non oidc-signing-junk page they were on. Not 100% sure this is flawless but it seems to solve the issue of wanting a dynamic back button that doesn't fall over on the oidc page (like the browser back sometimes does). Will show and tell soon.

#201 - I just reduced possible number of search results to speed things up and hopefully resolve this lag induced issue. If it persists I'll try increasing the search debounce too.

#202 - Showed batch number on search results table.


src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -302,6 +302,7 @@ private void BuildWorkOrders(ModelBuilder builder)
             q.Property(e => e.WorkStationCode).HasColumnName("WORK_STATION_CODE").HasMaxLength(16);
             q.Property(e => e.Outstanding).HasColumnName("OUTSTANDING").HasMaxLength(1);
             q.Property(e => e.ZoneName).HasColumnName("ZONE_NAME").HasMaxLength(20);
+            q.Property(e => e.BatchNotes).HasColumnName("QUALITY_ISSUES");
```
atcampbell: What's the reason for calling this BatchNotes instead of QualityIssues?

src/Service.Host/client/src/components/worksOrders/WorksOrderBatchNotes.js
```diff
@@ -31,7 +32,7 @@ function WorksOrdersBatchNotes({ items, fetchItems, loading, clearSearch, histor
                         table={table}
                         columnNames={columnNames}
                         fetchItems={fetchItems}
-                        clearSearch={clearSearch}
+                        clearSearch={() => {}}
```
atcampbell: I think clearSearch should be removed as a required prop type

src/Service.Host/client/src/middleware/previousLocation.js
```diff
@@ -0,0 +1,16 @@
+let prevPathname = '';
```
atcampbell: Is it worth adding this stuff to the shared library?

src/Domain.LinnApps/WorksOrders/WorksOrder.cs
```diff
@@ -52,21 +52,24 @@ public class WorksOrder
...
-        public void CancelWorksOrder(int? cancelledBy, string reasonCancelled)
+        public string BatchNotes { get; set; }
+
+        public void UpdateWorksOrder(int quantity, string batchNotes, int? cancelledBy, string reasonCancelled)
         {
-            if (cancelledBy == null || string.IsNullOrEmpty(reasonCancelled))
+            if (reasonCancelled != null)
             {
-                throw new InvalidWorksOrderException("You must provide a user number and reason when cancelling a works order");
+                if (cancelledBy == null || string.IsNullOrEmpty(reasonCancelled))
```
atcampbell: cancelledBy is always being set in the works order service so I think this will always try throw when trying to update a WO without a reasonCancelled. Maybe just check if reasonCancelled is not null?

lewisrenfrew: Seems to work ok... Where do you reckon cancelled by is being set?

src/Persistence.LinnApps/ServiceDbContext.cs
lewisrenfrew: just seems to be what it's called in practice... I thought this seemed the least confusing place to rename it other than renaming the actual database column

src/Service.Host/client/src/components/worksOrders/WorksOrderBatchNotes.js
lewisrenfrew: yeah, was thinking that too

src/Service.Host/client/src/middleware/previousLocation.js
lewisrenfrew: maybe eventually... I think I'll show and tell it first and see what the verdict is. Not sure it's a perfect solution.

src/Domain.LinnApps/WorksOrders/WorksOrder.cs
atcampbell: Is it set in WorksOrderService.cs? I could just be misreading things though.

lewisrenfrew: The cancelledBy field of the resource is set in the worksOrder service, but it only every gets put onto the domain object if a reason is provided. 


The UpdateWorksOrder method now looks like: 
```
public void UpdateWorksOrder(int quantity, string batchNotes, int? cancelledBy, string reasonCancelled)
        {
            if (reasonCancelled != null)
            {
                if (cancelledBy == null || string.IsNullOrEmpty(reasonCancelled))
                {
                    throw new InvalidWorksOrderException("You must provide a user number and reason when cancelling a works order");
                }

                this.CancelledBy = cancelledBy;
                this.ReasonCancelled = reasonCancelled;
                this.DateCancelled = DateTime.UtcNow;
            }

            this.Quantity = quantity;
            this.BatchNotes = batchNotes;
        }
    }
```

So I think things are ok.

lewisrenfrew: It should be on sys by now, so feel free to try and break it and let me know if you do.

atcampbell: Ahh yeah

PR #221, opened 2020-01-07
Manufacturing routes - make fields optional

iaincrawf:


iaincrawf:
> Cool. I take it they're not required at database level?

Both nullable in the db and tested it locally so should be fine - I've actually already come across some existing records saved without those fields, think it was a mistake making them required in the first place.

PR #224, opened 2020-01-08
set default start date to start of a week

richardiphillips:
tiny tweak so I can let people test this.


atcampbell:
> I've put a function into the shared library getWeekStartDate (in utilities/dateUtilities) that can do this.

Which gets the start of a linn week from a date

richardiphillips:
> > I've put a function into the shared library getWeekStartDate (in utilities/dateUtilities) that can do this.
> 
> Which gets the start of a linn week from a date

Thanks. I'll switch to that in my current branch

PR #241, opened 2020-01-15
Builtthisweek

### LINKED ISSUE #232: Osr Built this week drilldown
replacement for spud btw detail

andrewfraser73:
Simple built this week drill down from OSR using the report components

Resolves #232

### DIFF
--- a/src/Domain.LinnApps/BuiltThisWeekStatistic.cs
+++ b/src/Domain.LinnApps/BuiltThisWeekStatistic.cs
@@ -1,0 +1,19 @@
+﻿namespace Linn.Production.Domain.LinnApps
+{
+    public class BuiltThisWeekStatistic
+    {
+        public string CitCode { get; set; }
+
+        public string CitName { get; set; }
+
+        public string PartNumber { get; set; }
+
+        public string Description { get; set; }
+
+        public int BuiltThisWeek { get; set; }
+
+        public decimal Value { get; set; }
+
+        public double Days { get; set; }
+    }
+}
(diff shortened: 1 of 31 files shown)


src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -142,6 +142,8 @@ public class ServiceDbContext : DbContext
 
         private DbQuery<PtlMaster> PtlMasterSet { get; set; }
 
+        public DbQuery<BuiltThisWeekStatistic> BuiltThisWeekStatistics { get; set; }
```
richardiphillips: public should be before the privates. Can we get rid of those private dbquery entries now we've got the single record repo?

src/Resources/BtwRequestResource.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Linn.Production.Resources
+{
+    public class BtwRequestResource {
+        public string CitCode { get; set; }
```
richardiphillips: could just be cit code request resource for reuseability?

tests/Integration/Service.Tests/BuiltThisWeekModuleSpecs/ContextBase.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
richardiphillips: weird usings

src/Domain.LinnApps/BuiltThisWeekStatistic.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Linn.Production.Domain.LinnApps
```
richardiphillips: there's a folder for view models in the domain. Perhaps this should go in it

src/Persistence.LinnApps/ServiceDbContext.cs
andrewfraser73: yes we should probably rewrite these - will raise an issue.

PR #244, opened 2020-01-15
Ate test ut 2

lewisrenfrew:


src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
```diff
@@ -286,4 +286,4 @@ ManufacturingRoute.defaultProps = {
     manufacturingSkills: []
 };
 
-export default ManufacturingRoute;
+export default ManufacturingRoute;
```
iaincrawf: not an issue but I can't see what it sees as the difference here hahah

lewisrenfrew: I removed the carriage return at the end of the line, will add back

PR #247, opened 2020-01-16
Copy function (create trigger level)

iaincrawf:
#239


src/Service.Host/client/src/actions/triggersActions.js
```diff
@@ -0,0 +1,12 @@
+﻿import { FetchApiActions } from '@linn-it/linn-form-components-library';
+import { productionTriggerLevels as actionTypes } from './index';
+import * as itemTypes from '../itemTypes';
+import config from '../config';
+
```
atcampbell: Should this file be productionTriggerLevelsActions?

iaincrawf: That one does exist and uses UpdateApiActions, but I needed to also have FetchApiActions to use the typeahead search thing.. should I maybe call it something like productionTriggerLevelsSearchActions or put them in the same file?

PR #250, opened 2020-01-20
Osr deliver performance drill down

### LINKED ISSUE #235: osr Production Delivery Performance by CIT
OSR Production Delivery Performance by CIT
Has drill down to prod_del_perf_detail

andrewfraser73:
The delivery performance drill down and detail drill down (so its a meaty 52 files).

This uses the common.reporting for the reports but it does bring a couple of functions server side that were PL/SQL in the Oracle world

- Linn_Week_pack.start_week_date(date) and end_week_date(date).  Rather than write proxies for these I have written c# code that makes Saturday the first day of the week (as Linn decreed) rather than Microsoft's Monday "theory"

- Ptl Stat summary.  Rather than have two view one for the summary and one for the detail I just hoover up the stats and server side do the calcs using a PtlStatSummary class and c#.  The only tricky part is the percentile calculation which to be honest is a statistic probably ignored by the OSR userbase most of the time.

Resolves #235


src/Domain.LinnApps/Models/PtlStatPrioritySummary.cs
```diff
@@ -0,0 +1,85 @@
+﻿namespace Linn.Production.Domain.LinnApps.Models
+{
+    using System;
+    using System.ComponentModel.Design;
+    using System.Data.SqlTypes;
+    using System.Xml.Schema;
```
richardiphillips: xml schema?

src/Domain.LinnApps/Models/PtlStatPrioritySummary.cs
```diff
@@ -0,0 +1,85 @@
...
+        public decimal Percentile95 { get; set; }
+
+        public decimal PercBy5Day()
+        {
+            if (this.Triggers == 0)
+            {
+                return 0;
+            }
+
+            return Decimal.Round((((decimal)(this.Triggers - this.Gt5Day)) / (decimal)this.Triggers) * 100);
```
richardiphillips: stylecop prefers decimal to Decimal

src/Domain.LinnApps/Reports/DeliveryPeformanceReportService.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public ResultsModel GetDeliveryPerformanceDetail(string citCode, int priority)
+        {
+            var dates = this.GetDefaultDateRange();
+            var stats = this.ptlStatRepository.FilterBy(s =>
+                s.CitCode == citCode && s.PtlPriority == priority && s.DateCompleted >= dates.fromDate && s.DateCompleted <= dates.toDate).ToList();
+
+            var model = new ResultsModel();
+            model.ReportTitle = new NameModel("Production Delivery Performance " + dates.fromDate.ToString("dd-MMM-yy") + " - " + dates.toDate.ToString("dd-MMM-yy") + " Priority " + priority + " Cit " + citCode);
```
richardiphillips: would be nicer with string interpolation $""

PR #251, opened 2020-01-20
Manufacturing routes cycle time - save 1 decimal place

iaincrawf:
#252 
http://rndsupport.linn.co.uk/scp/tickets.php?id=12193


src/Domain.LinnApps/ManufacturingOperation.cs
```diff
@@ -44,7 +44,7 @@ private ManufacturingOperation()
 
         public int SetAndCleanTime { get; set; }
 
-        public int CycleTime { get; set; }
+        public double CycleTime { get; set; }
```
richardiphillips: I think decimal is better than double. 
Can use decimal.Round too

iaincrawf: Hadn't actually come across decimal before - consensus I can find online seems to be that it can handle bigger/more precise numbers and keeps it in decimal instead of binary so it's better for exact values, is that pretty much right? This data in column seems to just be a few minutes for most entries and we only care about 1 decimal place in it but I guess it's a good habit to get in to for calculation stuff

richardiphillips:
Yeah. It’s the exact value thing that does it for me.

PR #258, opened 2020-01-28
Label printer v1

iaincrawf:


src/Facade/Services/LabelPrintService.cs
```diff
@@ -67,9 +47,115 @@ public IResult<IEnumerable<IdAndName>> GetLabelTypes()
...
+                              {
+                                  { GeneralPurposeLabelTypes.Labels.PCNumbers, this.PrintPcNumbers(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.AddressLabel, this.PrintAddressLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.Small, this.PrintSmallLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.SmallWeeText, this.PrintSmallWeeTextLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.SmallBoldText, this.PrintSmallWeeBoldTextLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.GoodsInLabel, this.PrintGoodsInLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.LargeBigText, this.PrintLargeBigTextLabel(resource, dateTimeNow, printer) },
+                                  { GeneralPurposeLabelTypes.Labels.LargeWeeText, this.PrintLargeWeeTextLabel(resource, dateTimeNow, printer) }
```
richardiphillips: i know this isn't complete because this is really a PR for testing but for reference I think nearly all this code in the facade service should be moved to the domain.

PR #261, opened 2020-01-29
label printer updates

iaincrawf:
Logic moved into domain + some fixes. Should hopefully be able to get people to test this version and get feedback. Still needs the smiley face added


src/Domain.LinnApps/LabelPrintingService.cs
```diff
@@ -0,0 +1,137 @@
...
+            ILabelService labelService)
+        {
+            this.labelService = labelService;
+        }
+        
+        public LabelPrintResponse PrintLabel(LabelPrint resource)
+        {
+            var labelType = (GeneralPurposeLabelTypes.Labels)resource.LabelType;
+
+            var dateTimeNow = DateTime.Now.ToString("ddMMMyyyyHH''mm''ss");
```
richardiphillips: Doesn't matter but could just use iso format for simplicity?

src/Domain.LinnApps/LabelPrintingService.cs
```diff
@@ -0,0 +1,137 @@
...
+            var printer = ((LabelPrinters.Printers)resource.Printer).ToString();
+
+            var printMapper = new Dictionary<GeneralPurposeLabelTypes.Labels, Func<LabelPrint, string, string, LabelPrintResponse>>
+                              {
+                                  { GeneralPurposeLabelTypes.Labels.PCNumbers, this.PrintPcNumbers },
```
richardiphillips: This seems quite a complicated solution for printing a label but I'm sure it's very clever.

src/Domain.LinnApps/LabelPrintingService.cs
```diff
@@ -0,0 +1,137 @@
...
+                        printer,
+                        resource.Quantity,
+                        "c:\\lbl\\PCLabel.btw",
+                        pcNumber.ToString());
+                }
+
+                return new LabelPrintResponse(
+                        $"printed pc numbers {from} to {to}");
+            }
+            throw new ArgumentException("No PC number provided");
```
richardiphillips: Nice one. Looks like a good re-usable DomainException. Could do with a blank line after the if {}.

src/Facade/Services/LabelPrintService.cs
```diff
@@ -49,124 +49,39 @@ public IResult<IEnumerable<IdAndName>> GetLabelTypes()
...
-                + $"\"{resource.LinesForPrinting.Line4}\", \"{resource.LinesForPrinting.Line5}\", \"{resource.LinesForPrinting.Line6}\","
-                + $" \"{resource.LinesForPrinting.Line7}\"";
-
-            this.labelService.PrintLabel($"L1{dateTimeNow}", printer, resource.Quantity, "c:\\lbl\\genLargeLabel_1line.btw", data);
-
-            return new SuccessResult<LabelPrintResponse>(new LabelPrintResponse($"printed large (wee text) label{(resource.Quantity != 1 ? "s" : "")}"));
+            var printDetails = new LabelPrint
+                                   {
+                                       LabelType = resource.LabelType,
+                                       LinesForPrinting = new LabelPrintContents()
```
richardiphillips: Don't need the ()

src/Facade/Services/LabelPrintService.cs
```diff
@@ -49,124 +49,39 @@ public IResult<IEnumerable<IdAndName>> GetLabelTypes()
...
+                                                                  PartNumber = resource.LinesForPrinting.PartNumber,
+                                                                  Qty = resource.LinesForPrinting.Qty,
+                                                                  Initials = resource.LinesForPrinting.Initials,
+                                                                  Date = resource.LinesForPrinting.Date
+                                       },
+                                       Printer = resource.Printer,
+                                       Quantity = resource.Quantity
+                                   };
+
+            var result = this.labelPrintingService.PrintLabel(printDetails);
```
richardiphillips: Much nicer facade

src/Domain.LinnApps/LabelPrintingService.cs
iaincrawf: Separating into a method for each type of label we're printing seemed an ok idea to me, so it was just for how to get from the type of label into the right method to print that type. There likely is a simpler way or maybe a switch statement would've been ok, but I really just wanted something nicer/more readable (maybe I didn't manage that bit!) than 8 "else if" statements

iaincrawf:
Haven't got tests yet, think that's next on my list

> LGTM. Have you got tests for the domain service?

src/Domain.LinnApps/LabelPrintingService.cs
iaincrawf: That's .ToString("o") right? Just wasn't sure what would be allowed in the filename in terms of dashes and dots etc so played it safe, I can try it if you want!

PR #262, opened 2020-01-29
Fflag report

richardiphillips:


failed_parts_report_view.sql
```diff
@@ -0,0 +1,25 @@
...
+supp.supplier_name
+from v_tqms_fflagged f,
+parts p, suppliers supp,
+cits c, auth_user_name_view au
+where f.part_number =  p.part_number 
+and which_cit(p.part_number) = c.code (+) 
+and p.accounting_company = 'LINN'
+and f.created_by = au.user_number
+and p.preferred_supplier = supp.supplier_id (+)
+and f.being_verified = 'N'
```
lewisrenfrew: How/where is this sql used?

richardiphillips: It doesn't have to be used again. Just a record of the view I created for the report which in turn is a simplified version of the sql in the old report.

PR #277, opened 2020-02-05
Ate test and batch viewer fixes

lewisrenfrew:


src/Facade/Services/AteTestService.cs
```diff
@@ -67,6 +68,8 @@ protected override AteTest CreateFromResource(AteTestResource resource)
                                     Machine = detail.Machine,
                                     Shift = detail.Shift,
                                     SmtOrPcb = detail.SmtOrPcb,
+                                    PcbOperator = detail.PcbOperator != null ?
```
richardiphillips: if PcbOperator is int? then this might be nicer as PcbOperator.HasValue ? findbyid(PcbOperator.Value) : null

src/Facade/Services/WorksOrdersService.cs
```diff
@@ -91,24 +91,37 @@ public IResult<WorksOrder> UpdateWorksOrder(WorksOrderResource resource)
...
+                        .GetProperty(
+                            char.ToUpperInvariant(orderByDesc[0]) 
+                            + orderByDesc.Substring(1)).GetValue(w, null)); 
+            }
 
-            if (result.Count() > 1000)
+            if (limit != null)
             {
-                result = result.Take(1000);
+                result = result.Take((int)limit);
```
richardiphillips: again limit.Value reads more nicely in my opinion

PR #287, opened 2020-02-10
Label printing - readd castlemilk and maybe fix pc numbers

iaincrawf:


src/Domain.LinnApps/LabelPrinters.cs
```diff
@@ -14,6 +14,7 @@ public enum Printers
             [Display(Name = "Goods In Labels (Small)")] GISmall,
             [Display(Name = "Production Label Printer (Small)")] ProdLbl1,
             [Display(Name = "Production Label Printer (Large)")] ProdLbl2,
+            [Display(Name = "Castlemilk Labels")] Castlemilk,
```
andrewfraser73: now might be the change at least in the display name to rename Castlemilk to Metalwork as they no longer operate out of Castlemilk (although everyone understands what we mean by that)

Issue #108, opened 2019-09-20
Manufacturing Resources Utility - Cost field says its required but is blank for some entries


lewisrenfrew:
not sure what the desired behavior is here but at current its quite confusing

iaincrawf:
Description shouldn't really be required either, same with mfg routes. Will fix

iaincrawf:
oops routes is already sorted

PR #288, opened 2020-02-10
mfg resources - make cost & description not required

iaincrawf:
#108

### DIFF
--- a/src/Service.Host/client/src/components/manufacturingResources/ManufacturingResource.js
+++ b/src/Service.Host/client/src/components/manufacturingResources/ManufacturingResource.js
@@ -39,10 +39,8 @@ function ManufacturingResource({
     }, [item, prevManufacturingResource]);
 
     const resourceCodeInvalid = () => !manufacturingResource.resourceCode;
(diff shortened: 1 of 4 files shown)


src/Service.Host/client/src/components/manufacturingResources/ManufacturingResource.js
```diff
@@ -39,10 +39,8 @@ function ManufacturingResource({
     }, [item, prevManufacturingResource]);
 
     const resourceCodeInvalid = () => !manufacturingResource.resourceCode;
-    const descriptionInvalid = () => !manufacturingResource.description;
-    const costInvalid = () => !manufacturingResource.cost;
 
-    const inputInvalid = () => resourceCodeInvalid() || descriptionInvalid() || costInvalid();
+    const inputInvalid = () => resourceCodeInvalid();
```
atcampbell: Could you just use the resourceCodeInvalid variable rather than having these two?

iaincrawf: ah yeah definitely could good spot

PR #290, opened 2020-02-11
Plo ord issue front end

lewisrenfrew:


src/Facade/ResourceBuilders/AddressResourceBuilder.cs
```diff
@@ -21,7 +21,7 @@ public AddressResource Build(Address address)
                 Line3 = address.Line3,
                 Line4 = address.Line4,
                 PostCode = address.PostCode,
-                Country = address.Country,
+                Country = address.Country.Name,
```
iaincrawf: Some of the data is crap for addresses so I think we sometimes have null in the country column - not sure if we need just like `Country = address.Country?.name` or if we also need to make Country nullable but think we might need to check for it

src/Service.Host/client/src/actions/index.js
```diff
@@ -297,6 +297,11 @@ export const failedPartsReportActionTypes = makeReportActionTypes(
     reportTypes.failedPartsReport.actionType
 );
 
+export const issueSernosActionTypes = makeProcessActionTypes(processTypes.issueSernos.actionType);
+
+export const buildSernosActionTypes = makeProcessActionTypes(processTypes.buildSernos.actionType);
+
+export const purchaseOrderActionTypes = makeActionTypes(itemTypes.purchaseOrder.actionType);
 export const daysRequiredReportActionTypes = makeReportActionTypes(
```
iaincrawf: Will linting want an extra line in here?

src/Service.Host/client/src/reducers/index.js
```diff
@@ -209,6 +212,16 @@ const reducer = history =>
         worksOrderLabel,
         worksOrderLabels,
         workStations,
+        componentCounts,
+        labelPrint,
+        labelPrinters,
+        labelPrintTypes,
+        suppliers,
+        addresses,
```
iaincrawf: If you can be bothered I'm sure Richard or someone had sorted these all alphabetically but it obviously isn't essential!

src/Service.Host/client/src/actions/index.js
lewisrenfrew: no.. but consistency wants an extra line so will add it anyways

src/Facade/ResourceBuilders/AddressResourceBuilder.cs
lewisrenfrew: I think entity framework will be able to resolve the one to many relationship between countries and addresses ok like this

src/Domain.LinnApps/Country.cs
```diff
@@ -7,5 +7,7 @@ public class Country
         public string CountryCode { get; set; }
 
         public string Name { get; set; }
+
+        public IEnumerable<Address> Addresses { get; set; }
```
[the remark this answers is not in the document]
lewisrenfrew: I think entity framework will be able to resolve the one to many relationship between countries and addresses ok like this

PR #295, opened 2020-02-14
Update

lewisrenfrew:
- update all packages to latest non beta versions 

- including the linter, which got a bit more severe in it's newer versions. I tried to fix some the new warnings (and some of the old ones) although there's still quite a few remaining


src/Service.Host/client/src/components/productionTriggerLevels/TriggerLevel.js
```diff
@@ -22,6 +22,7 @@ import {
 } from '@linn-it/linn-form-components-library';
 import Page from '../../containers/Page';
 
+//eslint-disable-next-line react/jsx-props-no-spreading
 const Transition = React.forwardRef((props, ref) => <Slide direction="up" ref={ref} {...props} />);
```
lewisrenfrew: New linter doesn't like spreading of props into components: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

@iaincrawf What exactly is thie Slide component doing? Does it require access to every single prop? (git says you wrote this file) 

If all props are required I'll leave the ignore in... otherwise I'll just pass the specific required props.

src/Service.Host/client/src/components/reports/measures/LinkTabCell.js
```diff
@@ -21,7 +21,7 @@ const useStyles = makeStyles(theme => ({
 }));
 
 function LinkTabCell(props) {
-    const { children, href, hoverHref, setHoverHref, ...other } = props;
+    const { children, href, hoverHref, setHoverHref } = props;
     const classes = useStyles();
```
lewisrenfrew: @andrewfraser73 
New linter doesn't like spreading of props into components: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

I couldn't work out why LinkTabCell needs these ...other props so I just removed them to appease the linter. 

Looking in TabCell and LinkTabCell, it seems they only need access to the specific props passed to them. I can't see any other usages of  the ...other props.

Let me know if removing them will break anything? (Should be on sys soon if you want to check in real life)

src/Service.Host/client/src/components/reports/measures/TabCell.js
```diff
@@ -3,11 +3,11 @@ import TableCell from '@material-ui/core/TableCell';
 import PropTypes from 'prop-types';
 
 function TabCell(props) {
-    const { children, value, index, ...other } = props;
+    const { children, value, index } = props;
 
     if (index !== value) return null;
 
-    return <TableCell {...other}>{children}</TableCell>;
```
lewisrenfrew: @andrewfraser73  Does TableCell use any of these? seems unlikely since its not our code

src/Service.Host/client/src/components/reports/measures/LinkTabCell.js
andrewfraser73: LinkTabCell is just a wrapped Table Cell which means if you allow the other to spread then you can give it TableCell props as well without having to specify all the ones you might want to use in the prop definitions.

The linter might not like spreading but when you are making enhanced versions of other components it is useful.

Not sure whether I actually used this or not

src/Service.Host/client/src/components/reports/measures/TabCell.js
andrewfraser73: see previous response about LinkTabCell.  TabCell is just a cut down version of that

src/Service.Host/client/src/components/assemblyFails/AssemblyFail.js
```diff
@@ -265,12 +265,12 @@ function AssemblyFail({
                                         />
                                     </Grid>
                                     <Grid item xs={10} />
-                                </Fragment>
+                                </>
                             ) : (
-                                <Fragment />
+                                <></>
```
atcampbell: I'm not sure how I feel about the empty tags instead of Fragments. Maybe I just need to get used to it but I think things like this are a bit messy and `<>` isn't totally clear.

src/Service.Host/client/src/components/boardFailTypes/BoardFailTypes.js
```diff
@@ -57,14 +57,14 @@ const BoardFailTypes = ({ loading, itemError, history, items }) => {
     return (
         <Page>
             <Title text="Board Fail Types" />
-            {itemError && <ErrorCard errorMessage={itemError.statusText} />}
+            {itemError && <ErrorCard errorMessage={itemError?.statusText} />}
```
atcampbell: Don't think this `?` is needed as the error card is only rendered if itemError is truthy

src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
```diff
@@ -201,15 +201,15 @@ function ManufacturingRoute({
                 </Grid>
                 {itemErrors && (
                     <Grid item xs={12}>
-                        <ErrorCard errorMessage={itemErrors.statusText} />
+                        <ErrorCard errorMessage={itemErrors?.statusText} />
```
atcampbell: Unnecessary ?

src/Service.Host/client/src/components/manufacturingSkills/ManufacturingSkill.js
```diff
@@ -82,7 +82,7 @@ function ManufacturingSkill({
                 </Grid>
                 {itemError && (
                     <Grid item xs={12}>
-                        <ErrorCard errorMessage={itemError.statusText} />
+                        <ErrorCard errorMessage={itemError?.statusText} />
```
atcampbell: Unnecessary ?

src/Service.Host/client/src/components/partFails/PartFail.js
```diff
@@ -73,8 +73,8 @@ function PartFail({
         if (editStatus === 'create' && profile) {
             setPartFail(a => ({
                 ...a,
-                enteredBy: profile.employee.replace('/employees/', ''), // the current user
-                enteredByName: profile.name
+                enteredBy: profile?.employee.replace('/employees/', ''), // the current user
```
atcampbell: Don't think these ?s are needed either - already checking profile above

src/Service.Host/client/src/components/reports/ManufacturingCommitDateReport.js
```diff
@@ -114,9 +114,15 @@ const ManufacturingCommitDateReport = ({ reportData, loading, options }) => {
...
+    reportData: PropTypes.shape({
+        results: PropTypes.arrayOf(PropTypes.shape({})),
+        totals: PropTypes.arrayOf(PropTypes.number),
+        incompleteLinesAnalysis: PropTypes.shape({
+            reportResults: PropTypes.arrayOf(PropTypes.shape)
+        })
+    }),
     loading: PropTypes.bool,
-    options: PropTypes.shape({})
+    options: PropTypes.shape({ date: PropTypes.instanceOf(Date) })
```
atcampbell: Is this checking that its a date object? Is it not an iso string here?

src/Service.Host/client/src/components/reports/measures/LinkTabCell.js
```diff
@@ -36,19 +35,22 @@ function LinkTabCell(props) {
...
             href={href}
+            index={index}
+            value={value}
             onMouseEnter={onMouseEnter}
             onMouseLeave={onMouseLeave}
-            {...other}
         >
             <Link component={RouterLink} to={href}>
-                {children}
+                <> {children} </>
```
atcampbell: Don't think these fragments are needed?

src/Service.Host/client/src/components/reports/triggers/FactListItem.js
```diff
@@ -27,32 +27,30 @@ function FactListItem({ children, header, secondary, avatar }) {
             return !open ? <ExpandLess /> : <ExpandMore />;
         }
 
-        return <Fragment />;
+        return <> </>;
```
atcampbell: Should this just return null if nothings being rendered?

src/Service.Host/client/src/components/reports/triggers/ProductionTriggers.js
```diff
@@ -75,11 +75,11 @@ function ProductionTriggers({ reportData, loading, cits, fetchTriggers, history,
                     {loading ? <Loading /> : ''}
                 </Grid>
                 {itemError ? (
-                    <ErrorCard errorMessage={itemError.details?.message} />
+                    <ErrorCard errorMessage={itemError?.details?.message} />
```
atcampbell: ? thing again - already know it is there by this point

src/Service.Host/client/src/components/serialNumberReissue/SerialNumberReissue.js
```diff
@@ -93,7 +93,7 @@ function SerialNumberReissue({
         if (reissuedSerialNumber) {
             setSelectedSerialNumber(sernos => ({
                 ...sernos,
-                newSerialNumber: reissuedSerialNumber.newSerialNumber
+                newSerialNumber: reissuedSerialNumber?.newSerialNumber
```
atcampbell: Think this ? isn't needed either -> checking above

src/Service.Host/postcss.config.js
```diff
@@ -1,12 +1,5 @@
...
-                '>1%',
-                'last 4 versions',
-                'Firefox ESR',
-                'not ie < 9'
-            ]
-        })
-    ]
-};
\ No newline at end of file
+﻿const autoprefixer = require('autoprefixer')({});
```
atcampbell: import syntax?

src/Service.Host/client/src/components/assemblyFails/AssemblyFail.js
lewisrenfrew: Yeah... it's a bit weird isn't it. I'm not sure why the linter prefers the shorthand for fragments. Do you think I should Revert them all and add an ignore to our linter config?

src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
lewisrenfrew: possibly... I added these ?s because without them a 'statusText not declared in PropTypes' warning occurs... I could just add statusText to the PropTypes declaration to get round it but in some cases I just went for the ? because errors can come in a few different shapes

src/Service.Host/client/src/components/assemblyFails/AssemblyFail.js
atcampbell: I don't think I mind either way really! More just commenting on how it seems to look a bit awkward compared to a Fragment - happy to go with the linter though.

src/Service.Host/client/src/components/manufacturingRoutes/ManufacturingRoute.js
atcampbell: I think the ?s here are repeating the && check above so maybe adding the prop types is the better solution. I appreciate this is probably a bit nitpicky and not massively necessary...

lewisrenfrew: I think you're right - the ? is just a sneaky way of avoiding the PropType check so best to put it in

PR #297, opened 2020-02-17
fixed bug due to removal of pc numbers

iaincrawf:


iaincrawf:
Couldn't make a breaking test due to being unable to change the label type - it's in a dropdown component which makes life a real pain

PR #308, opened 2020-02-25
Automate mw timings

iaincrawf:
This doesn't have all the data locally so will be partly testing the table on app


src/Domain.LinnApps/Reports/MWTimingsReportService.cs
```diff
@@ -0,0 +1,92 @@
...
+                    results.SetGridTextValue(rowIndex, 4, entry.ItemArray[4].ToString()); //resource code
+                    results.SetGridTextValue(rowIndex, 5, entry.ItemArray[5].ToString()); //start
+                    results.SetGridTextValue(rowIndex, 6, entry.ItemArray[6].ToString()); //end
+                    results.SetGridTextValue(rowIndex, 7, entry.ItemArray[7].ToString()); //built by
+                    results.AddToGridValue(rowIndex, 8, (decimal)entry.ItemArray[8]);              //mins taken
+                    rowIndex++;
+                    totalMinsTakenForPart += (int)entry.ItemArray[8];
+                }
+
+                var mwBuild = mwBuildsList.Find(x => x.ItemArray[0] == partNumber); //this might need to be comparing 2 strings, not sure how smart it'll be
```
lewisrenfrew: - not too sure what this comment means but think you can axe it

src/IoC/ServiceModule.cs
```diff
@@ -56,6 +56,10 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<ShortageSummaryReportService>().As<IShortageSummaryReportService>();
             builder.RegisterType<ProductionMeasuresReportService>().As<IProductionMeasuresReportService>();
             builder.RegisterType<LabelPrintingService>().As<ILabelPrintingService>();
+            builder.RegisterType<MWTimingsReportService>().As<IMWTimingsReportService>();
+            builder.RegisterType<MWTimingsReportProxy>().As<IMWTimingsDatabaseReportService>();
+
+
```
lewisrenfrew: blank

src/Domain.LinnApps/Reports/MWTimingsReportService.cs
iaincrawf: Ah yeah it's a bit I haven't been able to test since there isn't any data on my dev db for half of it but I'll just take it out, I should know that's what's wrong if it doesn't work

PR #327, opened 2020-03-06
Works order ut marks fixes

lewisrenfrew:
- allow choice of cancelledBy as opposed to just using current user
- move print buttons to top
- more clear cancelaltion process, opens in modal


src/Service.Host/client/src/components/worksOrders/WorksOrder.js
```diff
@@ -89,12 +94,24 @@ function WorksOrder({
 
     const classes = useStyles();
 
+    const cancelledByOptions = () => {
+        const list = employees
```
atcampbell: Not important for now - but do you think it could be worth making this a selector in the shared library? I think there's a few places we're doing this.

lewisrenfrew: probably yeah

Issue #328, opened 2020-03-09
Think about back buttons (again)


lewisrenfrew:
I thought I had provided a good solution to the back button confusion by keeping a log of history in redux store so that back buttons could work out dynamically where to go if needed... however:

A problem arises when a user is coming from the old world, e.g.  

http://app.linn.co.uk/production/smt/smtwo.aspx

and then clicking through to a specific works order that now lives in the new world.

http://app.linn.co.uk/production/works-orders/903610

There is no way of knowing where the user wants to go back to since they have come from outside the realms of the redux store... The browser back button isn't currently a workaround because of the oidc callback thing.

PR #331, opened 2020-03-09
use window history api when user came from another app

lewisrenfrew:
- addresses #328 
- takes users back to external urls if that's where they came from :)
- I'll probably write a selector and move the event handler into the shared components at some point, but want to check it works in practise


lewisrenfrew:
> Might be a good one for the shared library

I think so... I am going to merge this and test it in the field for a bit before I commit to the idea though

PR #347, opened 2020-03-30
Use Parts table instead of Mech Part Source

atcampbell:


src/Domain.LinnApps/AuthorisedAction.cs
```diff
@@ -22,6 +22,9 @@ public class AuthorisedAction
 
         public const string ManufacturingTimings = "manufacturing-timings.admin";
 
+        // TODO remove this
```
richardiphillips: why not just remove it?

src/Domain.LinnApps/AuthorisedAction.cs
```diff
@@ -22,6 +22,9 @@ public class AuthorisedAction
...
         public const string PartCadInfoUpdate = "part-cad-info.admin";
+
```
richardiphillips: will there still be an ecit type authorisation. The benefit of having this as a separate app rather than just being in the parts app would be that these fields were only available to that team

atcampbell: Would it maybe be better to change this back to updating through a `/production/maintenance/part-cad-info/part-number` route? Then check for the partCadInfo privilege rather than a general part one?

(As opposed to this change which updated through `/production/maintenance/part-cad-info/part-number`)

richardiphillips: I don't mind but maybe that would be clearer since Lewis is currently making the parts update route in stores

atcampbell:
Need to update the parts table with LIBRARY_REF, LIBRARY_NAME, FOOTPRINT_REF and remove LIBRARY_REF and FOOTPRINT_REF from MECH_PART_SOURCE table. Holding off updating tables today.

PR #349, opened 2020-07-06
improve build plan detail ordering

atcampbell:


kylegordon:
Seems to contain ASCII. LGTM :+1:

PR #352, opened 2020-09-21
fix search button after create

atcampbell:


richardiphillips (approved):
why does it go straight into edit?

PR #353, opened 2020-09-29
remove authorisation requirment for production trigger level GET

atcampbell:


src/Service/Modules/ProductionTriggerLevelsModule.cs
```diff
@@ -127,7 +127,6 @@ private object GetProductionTriggerLevelsSettings()
 
         private object GetProductionTriggerLevel(string partNumber)
         {
-            this.RequiresAuthentication();
             var privileges = this.Context?.CurrentUser?.GetPrivileges().ToList();
```
richardiphillips, line 130: do we still need the privileges in this case?

richardiphillips (approved):

iaincrawf (approved):
oops think that'd have been me

src/Service/Modules/ProductionTriggerLevelsModule.cs
richardiphillips: i guess the facade is just expecting them because they're used when posting or something?

atcampbell: No, I'll take them out

atcampbell: Oh actually, we still need them for the response processor - its expecting to give back a `<ResponseModel<ProductionTriggerLevel>>` and we still need authorisation around updating / deleting PTLs

atcampbell: > i guess the facade is just expecting them because they're used when posting or something?

Oops, somehow missed this second comment

PR #354, opened 2020-10-08
Assy fails fixes

lewisrenfrew:
ticket 12969
- expand scope of search
- handle cases where dropdown values are no longer valid
- show circuit part number along with ref


andrewfraser73 (approved):
+1

PR #359, opened 2020-12-17
added new permission to allow updating of ptl description field 

iaincrawf:
http://rndsupport.linn.co.uk/scp/tickets.php?id=13331

### DIFF
--- a/src/Facade/ResourceBuilders/ProductionTriggerLevelResourceBuilder.cs
+++ b/src/Facade/ResourceBuilders/ProductionTriggerLevelResourceBuilder.cs
@@ -59,6 +59,11 @@
             {
                 yield return new LinkResource { Rel = "edit", Href = "/production/maintenance/production-trigger-levels/" };
             }
+
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.ProductionTriggerLevelUpdateDescription, model.Privileges))
+            {
+                yield return new LinkResource { Rel = "edit-description", Href = "/production/maintenance/production-trigger-levels/" };
+            }
         }
     }
 }
--- a/src/Facade/ResourceBuilders/ProductionTriggerLevelResourceBuilder.cs
+++ b/src/Facade/ResourceBuilders/ProductionTriggerLevelResourceBuilder.cs
@@ -57,12 +57,12 @@
 
             if (this.authorisationService.HasPermissionFor(AuthorisedAction.ProductionTriggerLevelUpdate, model.Privileges))
             {
-                yield return new LinkResource { Rel = "edit", Href = "/production/maintenance/production-trigger-levels/" };
+                yield return new LinkResource { Rel = "edit", Href = $"/production/maintenance/production-trigger-levels/{model.ResponseData.PartNumber}" };
             }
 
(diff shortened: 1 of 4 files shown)


atcampbell (approved):

src/Facade/ResourceBuilders/ProductionTriggerLevelResourceBuilder.cs
```diff
@@ -59,6 +59,11 @@ private IEnumerable<LinkResource> BuildLinks(ResponseModel<ProductionTriggerLeve
             {
                 yield return new LinkResource { Rel = "edit", Href = "/production/maintenance/production-trigger-levels/" };
             }
+
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.ProductionTriggerLevelUpdateDescription, model.Privileges))
+            {
+                yield return new LinkResource { Rel = "edit-description", Href = "/production/maintenance/production-trigger-levels/" };
+            }
         }
```
lewisrenfrew, line 67: I would probably do a permission check when the user actually goes to try and update the description (e.g. in the Update Nancy route). I assume you're attaching it to the resource as a link so that you can use that in the client to enable or disable a field? Remember a users privileges are already in the client as part of the user profile object in redux state so you probably don't need to use Links here.

iaincrawf: Yeah I thought about checking in the backend when they go to update, but this privilege is only for the description so other users who can edit but don't have this new privilege are still allowed to update everything else, just not the description. It would need to pull the existing ptl from the db, check if the description was being changed and then check this privilege if so and thought it probably wasn't worth it for updating a description. There's already the proper check for editing the whole ptl so it's just the tiny risk of someone who already can edit the rest of the ptl messing with the disabled thing on the description and editing that too. I kinda thought in the offchance that someone who could already edit the rest did manage to that we probably wouldn't care that much. Meant to put mention that in this pr, sorry!

Had forgotten about the privileges being there as part of the user and just used the way that was there already, do you think it's worth changing this form to check the user in redux for the future while I'm here or not that much benefit?

lewisrenfrew: It's probably personal preference.

 I like to keep hypermedia links meaningful in the sense that the 'rel' is a description of a linked resource or action and the 'href' is the url at which you can make GETs, PUTs, POSTs etc to in order to carry out whatever is described by the rel. 

In your case it's an action, namely editing description, and  I'd expect the href  to be the url at which I carry out that action. Like I'd naively assume making a PUT to that url would edit the description by this convention, or perhaps doing a GET to that href would take me to a page where I would edit the description.

We probably haven't stuck to this religiously though, and I've probably broken the convention imyself, so feel free to ignore.

richardiphillips: yeah. like you say i'm sure i've bent the rules too but definitely the edit rels should refer to the actual item to be edited.

iaincrawf: Cool, I hadn't done that properly when I made the form originally and just copy & pasted the existing edit link for the new permission. I've updated them both now to the actual edit url for the current ptl

PR #362, opened 2021-01-20
add clear labels

richardiphillips:


src/Domain.LinnApps/RemoteServices/IBartenderLabelPack.cs
```diff
@@ -2,8 +2,6 @@
 {
     public interface IBartenderLabelPack
     {
-        string ReturnPackageMessage();
```
iaincrawf, line 5: What was this method that's gone, just returning something saying that they've printed? Did we not use it?

iaincrawf (approved):
Looks good to me

src/Domain.LinnApps/RemoteServices/IBartenderLabelPack.cs
richardiphillips: it returns a status message from the internal state of the oracle package. we weren't using it and I don't fully trust that we always get the same state between calls to a package anyway.

PR #367, opened 2021-02-09
Log time created

lewisrenfrew:


andrewfraser73 (approved):
LGTM

richardiphillips:
Is the entered by field so that they don't have to log in properly? If we don't enforce that then do we need to bring horrible user codes back?

lewisrenfrew:
> Is the entered by field so that they don't have to log in properly? If we don't enforce that then do we need to bring horrible user codes back?
I guess so - the request was that the currently logged in user isn't always the person the part fail is marked as 'enteredBy'. Maybe I should have pushed back

richardiphillips:
yeah I don't know. We should definitely tell them they need to log in properly. they're responsible for anything that happens when they're logged in. if we're allowing this then we'll need to think about how much we care about security and traceability for stock transactions etc etc.

PR #368, opened 2021-04-08
fix cancel/back

lewisrenfrew:
@richardiphillips As far as I could see the works order form is already using the smartGoBack function. It seems to correctly take you back to the trigger levels report - if you click through from there to raise a WO, clicking back does actually take you back to the report.

I did fix a weird bug related to cancel button though, hence the PR, which might have been what Mark meant? 

I can catch up with him about it


richardiphillips (approved):

richardiphillips:
he didn't mention cancelling. he just said he raised a works order and the back button didn't take him back to the report. I'm sure he'll be able to give you an example if you drop him a teams

PR #370, opened 2021-04-30
allow dorik product labels

richardiphillips:


iaincrawf (approved):
got confused thinking someone was typing in doric but akudorik speakers makes much more sense

PR #372, opened 2021-05-13
Failed parts report options

lewisrenfrew:
- haven't written any new tests yet since I'm not 100% sure this is what is being asked for. Will write tests in a seperate pr once I have the OK from Stephen
- I tried to keep it as flexible as possible, so you can now order by date asc or desc, or order by part number, or filter part numbers using an asterisk wildcard

### DIFF
--- a/src/Domain.LinnApps/Reports/ProductionMeasuresReportService.cs
+++ b/src/Domain.LinnApps/Reports/ProductionMeasuresReportService.cs
@@ -52,7 +52,8 @@
 
             if (!string.IsNullOrEmpty(partNumber) || !string.IsNullOrEmpty(orderByDate))
             {
-                if (partNumber.Contains("*"))
+                columns.Add(new AxisDetailsModel("CIT", GridDisplayType.TextValue));
+                if (partNumber != null && partNumber.Contains("*"))
                 {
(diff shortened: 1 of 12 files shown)


src/Domain.LinnApps/Reports/ProductionMeasuresReportService.cs
```diff
@@ -48,6 +45,57 @@ public IEnumerable<ResultsModel> FailedPartsReport(string citCode)
...
+                    var partNumberPattern = Regex.Escape(partNumber).Replace("\\*", ".*?");
+                    var r = new Regex(partNumberPattern, RegexOptions.IgnoreCase);
+                    fails = fails.Where(x => r.IsMatch(x.PartNumber.ToUpper()));
+                }
+                else
+                {
+                    fails = fails.Where(f => string.IsNullOrEmpty(partNumber) || f.PartNumber == partNumber.ToUpper());
+                }
+
+                if (orderByDate != null && orderByDate.Equals("ASC"))
```
richardiphillips, line 67: do you need the null check? would if (orderByDate == "") not be enough?

richardiphillips (approved):

PR #379, opened 2021-12-16
Bom bom bom

lewisrenfrew:
test incoming


iaincrawf (approved):
So cool and so groovy when it goes bom bom bom (the loops took me a minute to read through but think I understand now, looks good)

lewisrenfrew:
> So cool and so groovy when it goes bom bom bom (the loops took me a minute to read through but think I understand now, looks good)

its a bit of a mind bender but maybe the simple test I just pushed provides some kind of visualisation at least

PR #380, opened 2022-01-21
Allow and respect invalid fault codes + error types

iaincrawf:
Factory want to streamline them, so added date invalid to fault codes and disabled the dropdown options for invalid ones on the part fail screen so that they can't be chosen anymore. This way Rachel or whoever can tidy the data up to however they want it
http://rndsupport.linn.co.uk/scp/tickets.php?id=15147


iaincrawf:
Wee bit more to do actually, forgot about the part fail details report

src/Service.Host/client/src/components/partFails/PartFail.js
```diff
@@ -310,16 +322,33 @@ function PartFail({
...
+                                                <Select
+                                                    value={partFail.faultCode}
+                                                    label="Fault Code"
+                                                    propertyName="faultCode"
+                                                    allowNoValue={creating()}
+                                                    onChange={handleFaultCodeChange}
+                                                    required
+                                                >
+                                                    {/* display valid options at the top so we don't need to scroll through old ones */}
+                                                    {faultCodes.map(e => {
```
lewisrenfrew, line 336: Shouldn't we just omit invalid ones altogether?

lewisrenfrew (approved):

src/Service.Host/client/src/components/partFails/PartFail.js
iaincrawf: I'm not sure we should, then nothing would display when you look at any part fail that's been created up until now - they're wanting to invalidate almost everything they currently use for error types and fault codes. Unless we were to convert all of the old data to new ones but not sure that'd be right either

lewisrenfrew: Oh yeah, because it's a dropdown. it still seems a bit off to allow them to save an invalid one. You could stick a check in the back end to make sure they don't try and do so.

richardiphillips (approved):
if we are going to display the invalid options then I think there should be something that throws an error if they're selected. 

If we display something in a dropdown it leads to this kind of problem. 
Ideally it should just be a text display (including types that are now invalid) and in create mode it would be a drop down with only currently valid entries but i know that's more work here

src/Service.Host/client/src/components/partFails/PartFail.js
iaincrawf: You mean if they re-save an existing one without changing the fault code/error type? I think they kinda should be able to edit an old one without having to pick a new error type and the disabled on the menuItems stops them from being able to select it from now on, but we can chat about it in IT dev or something
image

src/Service.Host/client/src/components/partFails/PartFail.js
```diff
@@ -332,16 +361,33 @@ function PartFail({
                                         </Grid>
                                         <Grid item xs={3} />
                                         <Grid item xs={3}>
-                                            <Dropdown
```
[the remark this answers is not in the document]
iaincrawf: I couldn't disable individual options with our dropdown like I can here, the disabled flag on a MenuItem stops them from even selecting it. Thought that was slightly better than trusting people to pick the right ones! Lewis suggested stopping from saving on the backend with invalid options too, but that would stop updating other fields on existing part fails unless they selected a new option and not sure that's right either. Want me to post on IT dev to chat about this one?
image

src/Service.Host/client/src/components/partFails/PartFail.js
richardiphillips: ah i see. That answers both my questions. 
It's a very reasonable work around. 
I'd still have a check in place that new ones can't be added with invalid codes in an ideal world but it looks like practically speaking it can't happen anyway.

PR #397, opened 2022-04-20
add works order message as display only to works orders

iaincrawf:
new table just with part_number and message on it for support request http://rndsupport.linn.co.uk/scp/tickets.php?id=15598

We'd need to do the inserts to add to table at moment, might just release this as a v1 and make a wee add/edit page for the works order messages in a second version

_**@ me I need to add the table to prod before putting this pr live!**_


src/Service.Host/client/src/components/worksOrders/WorksOrder.js
```diff
@@ -529,6 +537,34 @@ function WorksOrder({
...
+                                    disabled
+                                    inputProps={{ style: { color: 'purple' } }}
+                                    InputLabelProps={{ shrink: true }}
+                                    variant="outlined"
+                                    fullWidth
+                                />
+                            </Grid>
+                            <Grid item xs={4}>
+                                {/* <Button
+                                    className={classes.printButton}
```
richardiphillips, line 554: what's this comment

src/Domain.LinnApps/Part.cs
```diff
@@ -46,6 +46,8 @@ public class Part
 
         public IEnumerable<BomDetail> BomDetailsWithThisPart { get; set; }
 
+        public WorksOrderMessage WorksOrderMessage { get; set; }
```
richardiphillips, line 49: i don't think i'd add this to part. i'd just make a repo for it and update the table manually for now. up to you though

richardiphillips (approved):

src/Domain.LinnApps/Part.cs
richardiphillips: in fact even if it's on the part I think it should just be a field in that case rather than another table to join to.

src/Service.Host/client/src/components/worksOrders/WorksOrder.js
iaincrawf: was originally wanting to make a wee ut to add/edit these messages at the same time so this was a link to it but realised it'd take longer than I had yesterday. I'll just remove that wee bit

src/Domain.LinnApps/Part.cs
iaincrawf: I thought just .including it from the part was easier, it won't be able to update etc because it only includes it on part filterby or worksorder findbyid, then does just add it as a string field to the partResource. I could just unlink it and do separate repo calls to get a potential message if you prefer? It's not on the actual part table, just this table has a foreign key to part_number so I thought it was convenient to add it onto part object so I could .include it

iaincrawf: it could also easily just be a field on the part table, though it's already massive

richardiphillips: yeah parts table wouldn't be my first choice. i'd go repo option personally i think but entirely up to you

iaincrawf: I feel like I've made this worse, not sure this is what you meant - calling a new wo message domain service from the part resource builder & works order resource builder to get the message and stick it on the resources. Also won't build cause I seem to have been an idiot with registering something that I just can't spot

richardiphillips: i don't think it has to be anywhere on the part resource. i think the works order domain service could handle it and just add it as a field on the works order? Or the message table could have its own service endpoint and the client look it up? The first one would be quickest

iaincrawf: I was pulling it in with parts so that I could display it when you select a part when you're creating works orders too, so that Stewart would know that part had a wo message when he was creating works orders, but I guess that's probably not a big issue and I could just put it on the works orders for viewing existing ones. Yeah didn't want to have the client look it up

src/Facade/ResourceBuilders/PartsResourceBuilder.cs
```diff
@@ -7,15 +7,16 @@
...
 
     public class PartsResourceBuilder : IResourceBuilder<ResponseModel<IEnumerable<Part>>>
     {
         private readonly PartResourceBuilder partResourceBuilder;
 
-        public PartsResourceBuilder(IAuthorisationService authorisationService)
+        public PartsResourceBuilder(IAuthorisationService authorisationService, IWorksOrderMessageService worksOrderMessageService)
         {
-            this.partResourceBuilder = new PartResourceBuilder(authorisationService);
+            this.partResourceBuilder = new PartResourceBuilder(authorisationService, worksOrderMessageService);
```
lewisrenfrew, line 19: this won't work I don't think, since this will happen at app start up and the worksOrderMessageService depends on an IRepository which only has instance per request autofac scope

lewisrenfrew: I think thats why the healthcheck is failing and the stacks got stuck

PR #400, opened 2022-05-16
PARTS FAIL DETAIL REPORT swap from week to day

RossStewart:
http://rndsupport.linn.co.uk/scp/tickets.php?id=15713

"The “ASSEMBLY FAILS DETAILS REPORT” can be ran from one day to another day, which is great. The “PARTS FAIL DETAIL REPORT” can only be ran week number to week number. I would like both reports to be ran Day to Day rather than week to week.

 

Matt"

### DIFF
--- a/src/Domain.LinnApps/Reports/PartsReportService.cs
+++ b/src/Domain.LinnApps/Reports/PartsReportService.cs
@@ -52,7 +52,7 @@
             var fromDateParse = DateTime.Parse(fromDate);
             var toDateParse = DateTime.Parse(toDate);
 
-            var fails = this.partFailLogRepository.FilterBy(p => p.DateCreated > fromDateParse && p.DateCreated < toDateParse);
+            var fails = this.partFailLogRepository.FilterBy(p => p.DateCreated >= fromDateParse && p.DateCreated <= toDateParse);
 
             var purchaseOrders =
                 this.purchaseOrderRepository.FilterBy(p => fails.Any(f => f.PurchaseOrderNumber == p.OrderNumber));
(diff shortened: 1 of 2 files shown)


src/Domain.LinnApps/Reports/PartsReportService.cs
```diff
@@ -42,17 +42,17 @@ public class PartsReportService : IPartsReportService
...
             string partNumber,
             string department)
         {
-            var fromDate = DateTime.Parse(fromWeek);
-            var toDate = DateTime.Parse(toWeek);
+            var fromDateParse = DateTime.Parse(fromDate);
+            var toDateParse = DateTime.Parse(toDate);
 
-            var fails = this.partFailLogRepository.FilterBy(p => p.DateCreated > fromDate && p.DateCreated < toDate);
+            var fails = this.partFailLogRepository.FilterBy(p => p.DateCreated > fromDateParse && p.DateCreated < toDateParse);
```
lewisrenfrew, line 55: should probably do >= and <= in the filterBy expression to include the date limits

src/Domain.LinnApps/Reports/PartsReportService.cs
```diff
@@ -94,7 +94,7 @@ public class PartsReportService : IPartsReportService
             var model = new ResultsModel
                             {
                                 ReportTitle = new NameModel(
-                                    $"Part Fail - Details for weeks {this.linnWeekPack.Wwsyy(fromDate)} - {this.linnWeekPack.Wwsyy(toDate)}")
+                                    $"Part Fail - Details for dates {fromDateParse.ToShortDateString()} - {toDateParse.ToShortDateString()}")
```
lewisrenfrew, line 97: you can explicitly set the format in these toString() calls to hopefully fix your test thats failing

lewisrenfrew (approved):

PR #436, opened 2023-10-19
actually handle issue sernos errors properly

lewisrenfrew:
and return a 400 with a useful error message


richardiphillips (approved):
nice. i take it the client already does all the display error business?

lewisrenfrew:
> nice. i take it the client already does all the display error business?

yeah pretty much

PR #445, opened 2025-04-24
Invalid skills/routes

RossStewart:
http://rndsupport.linn.co.uk/scp/tickets.php?id=20229

soft delete for Alison to make use of manufacturing skills/routes + small tidy/style change

### DIFF
--- a/src/Service.Host/ResponseProcessors/AddressResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AddressResponseProcessor.cs
@@ -1,0 +1,14 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps;
+
+    public class AddressResponseProcessor : JsonResponseProcessor<Address>
+    {
+        public AddressResponseProcessor(IResourceBuilder<Address> resourceBuilder)
+            : base(resourceBuilder, "addresses", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AddressesResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AddressesResponseProcessor.cs
@@ -1,0 +1,16 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps;
+
+    public class AddressesResponseProcessor : JsonResponseProcessor<IEnumerable<Address>>
+    {
+        public AddressesResponseProcessor(IResourceBuilder<IEnumerable<Address>> resourceBuilder)
+            : base(resourceBuilder, "address", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AssemblyFailFaultCodeResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AssemblyFailFaultCodeResponseProcessor.cs
@@ -1,0 +1,14 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.Measures;
+
+    public class AssemblyFailFaultCodeResponseProcessor : JsonResponseProcessor<AssemblyFailFaultCode>
+    {
+        public AssemblyFailFaultCodeResponseProcessor(IResourceBuilder<AssemblyFailFaultCode> resourceBuilder)
+            : base(resourceBuilder, "assembly-fail-fault-code", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AssemblyFailFaultCodesResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AssemblyFailFaultCodesResponseProcessor.cs
@@ -1,0 +1,16 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.Measures;
+
+    public class AssemblyFailFaultCodesResponseProcessor : JsonResponseProcessor<IEnumerable<AssemblyFailFaultCode>>
+    {
+        public AssemblyFailFaultCodesResponseProcessor(IResourceBuilder<IEnumerable<AssemblyFailFaultCode>> resourceBuilder)
+            : base(resourceBuilder, "assembly-fail-fault-codes", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AssemblyFailResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AssemblyFailResponseProcessor.cs
@@ -1,0 +1,14 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.Measures;
+
+    public class AssemblyFailResponseProcessor : JsonResponseProcessor<AssemblyFail>
+    {
+        public AssemblyFailResponseProcessor(IResourceBuilder<AssemblyFail> resourceBuilder)
+            : base(resourceBuilder, "assembly-fails-fault-codes", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AssemblyFailsResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AssemblyFailsResponseProcessor.cs
@@ -1,0 +1,16 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.Measures;
+
+    public class AssemblyFailsResponseProcessor : JsonResponseProcessor<IEnumerable<AssemblyFail>>
+    {
+    public AssemblyFailsResponseProcessor(IResourceBuilder<IEnumerable<AssemblyFail>> resourceBuilder)
+        : base(resourceBuilder, "assembly-fails", 1)
+    {
+    }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AteFaultCodeResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AteFaultCodeResponseProcessor.cs
@@ -1,0 +1,14 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.ATE;
+
+    public class AteFaultCodeResponseProcessor : JsonResponseProcessor<AteFaultCode>
+    {
+        public AteFaultCodeResponseProcessor(IResourceBuilder<AteFaultCode> resourceBuilder)
+            : base(resourceBuilder, "ate-fault-code", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AteFaultCodesResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AteFaultCodesResponseProcessor.cs
@@ -1,0 +1,16 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.ATE;
+
+    public class AteFaultCodesResponseProcessor : JsonResponseProcessor<IEnumerable<AteFaultCode>>
+    {
+        public AteFaultCodesResponseProcessor(IResourceBuilder<IEnumerable<AteFaultCode>> resourceBuilder)
+            : base(resourceBuilder, "ate-fault-codes", 1)
+        {
+        }
+    }
+}
--- a/src/Service.Host/ResponseProcessors/AteTestResponseProcessor.cs
+++ b/src/Service.Host/ResponseProcessors/AteTestResponseProcessor.cs
@@ -1,0 +1,14 @@
+﻿namespace Linn.Production.Service.ResponseProcessors
+{
+    using Linn.Common.Facade;
+    using Linn.Common.Nancy.Facade;
+    using Linn.Production.Domain.LinnApps.ATE;
+
+    public class AteTestResponseProcessor : JsonResponseProcessor<AteTest>
+    {
+        public AteTestResponseProcessor(IResourceBuilder<AteTest> resourceBuilder)
+            : base(resourceBuilder, "ate-test", 1)
+        {
+        }
+    }
+}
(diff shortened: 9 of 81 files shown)


src/Facade/Services/IManufacturingResourceFacadeService.cs
```diff
@@ -0,0 +1,13 @@
...
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Production.Domain.LinnApps;
+    using Linn.Production.Resources;
+
+    public interface IManufacturingResourceFacadeService : IFacadeFilterService<ManufacturingResource, string, ManufacturingResourceResource, ManufacturingResourceResource, ManufacturingResourceResource>
+    {
+        IResult<IEnumerable<ManufacturingResource>> GetValid();
+    }
```
lewisrenfrew, line 12: why do we need a whole new interface for this i.e. isnt this just a .FilterBy()?

src/Facade/Services/ManufacturingResourceFacadeService.cs
```diff
@@ -0,0 +1,61 @@
...
+            ITransactionManager transactionManager)
+            : base(manufacturingResourceRepository, transactionManager)
+        {
+            this.manufacturingResourceRepository = manufacturingResourceRepository;
+        }
+
+        public IResult<IEnumerable<ManufacturingResource>> GetValid()
+        {
+            return new SuccessResult<IEnumerable<ManufacturingResource>>(
+                this.manufacturingResourceRepository.FilterBy(c => !c.DateInvalid.HasValue));
```
lewisrenfrew, line 26: yeah this is just a filter by - don't need to introduce the complexity of a whole new interface to support it unless I'm missing something

src/Facade/Services/IManufacturingSkillFacadeService.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Production.Domain.LinnApps;
+    using Linn.Production.Resources;
+
+    public interface IManufacturingSkillFacadeService : IFacadeFilterService<ManufacturingSkill, string, ManufacturingSkillResource, ManufacturingSkillResource, ManufacturingSkillResource>
+    {
+        IResult<IEnumerable<ManufacturingSkill>> GetValid();
```
lewisrenfrew, line 11: as before

src/Domain.LinnApps/ManufacturingResource.cs
```diff
@@ -1,18 +1,15 @@
 ﻿namespace Linn.Production.Domain.LinnApps
 {
+    using System;
+
     public class ManufacturingResource
     {
-        public ManufacturingResource(string resourceCode, string description, double? cost)
```
richardiphillips, line 5: why are you killing these constructors? Perhaps cost could be taken out but you'd never want a resource without a code and description

src/Domain.LinnApps/ManufacturingSkill.cs
```diff
@@ -1,18 +1,15 @@
 ﻿namespace Linn.Production.Domain.LinnApps
 {
+    using System;
+
     public class ManufacturingSkill
     {
-        public ManufacturingSkill(string skillCode, string description, int? hourlyRate)
```
richardiphillips, line 5: as above

src/Service.Host/client/src/components/manufacturingSkills/ManufacturingSkill.js
```diff
@@ -70,6 +72,10 @@ function ManufacturingSkill({
         setManufacturingSkill({ ...manufacturingSkill, [propertyName]: newValue });
     };
 
+    const makeInvalid = () => {
+        handleFieldChange('dateInvalid', new Date());
+    };
+
```
lewisrenfrew, line 78: i would just reuse handleFieldChange for this, passing 'dateInvalid' and 'new date()' - no need for a seperate function

src/Service.Host/client/src/components/manufacturingResources/ManufacturingResource.js
```diff
@@ -66,6 +68,10 @@ function ManufacturingResource({
         setManufacturingResource({ ...manufacturingResource, [propertyName]: newValue });
     };
 
+    const makeInvalid = () => {
+        handleFieldChange('dateInvalid', new Date());
+    };
```
lewisrenfrew, line 73: again, wouldnt bother with a seperate function for this since handleFieldChange can accommodate

richardiphillips (commented):
agree that we don't need those interfaces. like the rename of the facade services.

src/Service/Modules/ManufacturingResourceModule.cs
```diff
@@ -2,29 +2,28 @@
...
-            this.manufacturingResourceService = manufacturingResourceService;
+            this.manufacturingResourceFacadeService = manufacturingResourceFacadeService;
             this.Get("/production/resources/manufacturing-resources/{resourceCode*}", parameters => this.GetManufacturingResourceById(parameters.resourceCode));
-            this.Get("/production/resources/manufacturing-resources", _ => this.GetAllManufacturingResources());
+            this.Get("/production/resources/manufacturing-resources", _ => this.GetManufacturingResources());
             this.Put("/production/resources/manufacturing-resources/{resourceCode*}", parameters => this.UpdateManufacturingResource(parameters.resourceCode));
             this.Post("/production/resources/manufacturing-resources", _ => this.AddManufacturingResource());
         }
 
-        private object GetAllManufacturingResources()
```
lewisrenfrew, line 25: I would be tempted to leave this in and create a new endpoint for your FilterBy... Might be nice to keep GetAll at the API level if they every wanted it for reporting purposes (i.e. just run a postman get all so you can see all including the dates they were invalidated).  Equally this might never be useful... so feel free to ignore. Or maybe float the idea to Alison and see if she might ever want such information

src/Domain.LinnApps/ManufacturingResource.cs
richardiphillips: Dunno if my comments made it through yesterday but you need to take dateinvalid out of the constructor. it should never be set at create time

richardiphillips: i'd change all doubles to decimal too.

src/Domain.LinnApps/ManufacturingResource.cs
```diff
@@ -1,18 +1,23 @@
...
+        public ManufacturingResource(string resourceCode, string description, double? cost, DateTime? dateInvalid)
         {
             this.ResourceCode = resourceCode;
             this.Description = description;
             this.Cost = cost;
+            this.DateInvalid = dateInvalid;
```
lewisrenfrew, line 12: I wouldnt pass dateInvalid in the constructor - seems odd that you'd want to create these in an 'invalid' state

src/Facade/Services/ManufacturingResourceFacadeService.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        protected override Expression<Func<ManufacturingResource, bool>> SearchExpression(string searchTerm)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override Expression<Func<ManufacturingResource, bool>> FilterExpression(ManufacturingResourceResource searchTerms)
+        {
+            return m => !m.DateInvalid.HasValue;
```
lewisrenfrew, line 48: this is a bit hardcode-y for a FilterExpression. it would be much better and more flexible to allow the calling code to specify whether it wants to include invalid resources or not

src/Facade/Services/ManufacturingSkillFacadeService.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        protected override Expression<Func<ManufacturingSkill, bool>> SearchExpression(string searchTerm)
+        {
+            throw new NotImplementedException();
+        }
+
+        protected override Expression<Func<ManufacturingSkill, bool>> FilterExpression(ManufacturingSkillResource searchTerms)
+        {
+            return m => !m.DateInvalid.HasValue;
```
lewisrenfrew, line 50: same again, I'd let the calling code specify its Filter conditions

src/Service.Host/client/src/components/__tests__/manufacturingResource.test.js
```diff
@@ -12,7 +12,8 @@ const setEditStatusMock = jest.fn();
 const manufacturingResource = {
     resourceCode: 'TESTCODE1',
     description: 'Descripticon',
-    cost: 10.55
+    cost: 10.55,
+    dateInvalid: new Date().toISOString()
```
lewisrenfrew, line 16: Why do we want test data to be marked as invalid, i.e. have a non-null dateInvalid, initially? I think this would only make sense in a test scenario where you wanted to test whether the user could sucessfully revalidate the the thing, i.e. remove a dateInalid where it was initially present, which we aren't doing here

src/Service.Host/client/src/components/__tests__/manufacturingResources.test.js
```diff
@@ -10,12 +10,14 @@ const manufacturingResources = [
     {
         resourceCode: 'TESTCODE1',
         description: 'Descripticon',
-        cost: 10
+        cost: 10,
+        dateInvalid: new Date().toISOString()
```
lewisrenfrew, line 14: Isn't the idea that this 'show all' page only shows valid ones? If so then I wouldn't make the test data consist of invalid onces, just for logical consistency

src/Service/Modules/ManufacturingResourceModule.cs
```diff
@@ -2,29 +2,40 @@
...
 
-        public ManufacturingResourceModule(
-            IFacadeService<ManufacturingResource, string, ManufacturingResourceResource, ManufacturingResourceResource> manufacturingResourceService)
+        public ManufacturingResourceModule(IFacadeFilterService<ManufacturingResource, string, ManufacturingResourceResource, ManufacturingResourceResource, ManufacturingResourceResource> manufacturingResourceFacadeService)
         {
-            this.manufacturingResourceService = manufacturingResourceService;
+            this.manufacturingResourceFacadeService = manufacturingResourceFacadeService;
             this.Get("/production/resources/manufacturing-resources/{resourceCode*}", parameters => this.GetManufacturingResourceById(parameters.resourceCode));
-            this.Get("/production/resources/manufacturing-resources", _ => this.GetAllManufacturingResources());
+            this.Get("/production/resources/manufacturing-resources/all", _ => this.GetAllManufacturingResources());
```
lewisrenfrew, line 19: conventionally we support getting all and searching via the same endpoint but decide what to do based on whether or not the client passes search parameters

src/Service/Modules/ManufacturingResourceModule.cs
```diff
@@ -2,29 +2,40 @@
...
+            var result = this.manufacturingResourceFacadeService.GetAll();
+            return this.Negotiate
+                .WithModel(result)
+                .WithMediaRangeModel("text/html", ApplicationSettings.Get)
+                .WithView("Index");
+        }
+
+        private object GetManufacturingResources()
+        {
+            var resource = this.Bind<ManufacturingResourceResource>();
```
lewisrenfrew, line 36: this resource  is just empty right? (since you are just hardcoding your FilterExpression and disregarding any client parameters anyway).  That seems a bit incorrect and supports the above comment - I think you should make this work in the standard correct way, i.e.

one /manufacturing-resources endpoint to retrieve the collection. The client can then specify its filter options as query parameters, something like  'includeInvalid=True/False" that you can use to decide whether to filter the collection or not

src/Service/Modules/ManufacturingSkillsModule.cs
```diff
@@ -10,20 +11,33 @@
...
 
-        public ManufacturingSkillsModule(IFacadeService<ManufacturingSkill, string, ManufacturingSkillResource, ManufacturingSkillResource> manufacturingSkillService)
+        public ManufacturingSkillsModule(IFacadeFilterService<ManufacturingSkill, string, ManufacturingSkillResource, ManufacturingSkillResource, ManufacturingSkillResource> manufacturingSkillFacadeService)
         {
-            this.manufacturingSkillService = manufacturingSkillService;
-            this.Get("/production/resources/manufacturing-skills", _ => this.GetAll());
+            this.manufacturingSkillFacadeService = manufacturingSkillFacadeService;
+
+            this.Get("/production/resources/manufacturing-skills/all", _ => this.GetAll());
+            this.Get("/production/resources/manufacturing-skills", _ => this.GetManufacturingSkills());
```
lewisrenfrew, line 21: same for this module. Much cleaner API design to properly user query parameters to specify how to filter the collections

tests/Integration/Service.Tests/ManufacturingResourcesModuleSpecs/WhenGettingValid.cs
```diff
@@ -1,24 +1,33 @@
...
         [SetUp]
         public void SetUp()
         {
-            var a = new ManufacturingResource("a", "desc", 15);
-            var b = new ManufacturingResource("b", "desc", 17);
-            this.ManufacturingResourceService.GetAll()
+            var a = new ManufacturingResource("a", "desc", 15, null);
+            var b = new ManufacturingResource("b", "desc", 17, null);
+            var c = new ManufacturingResource("c", "desc", 17, DateTime.Today);
+            var d = new ManufacturingResource("d", "desc", 17, DateTime.Today);
```
lewisrenfrew, line 28: these two 'invalid' resources are pointless (and possibly misleading) as test data - I can see were you were going with this since the general idea is filtering out invalid entries, but since you are mocking out your .FilterBy return value anyway i think this test data is misleading in that it suggests that the test  actually cares about/tests the filtering logic (which is untestable in this tests context since it runs in the repository which is being mocked out).

aside from that rant, c and d will be unused variables anyway and so should be deleted regardless

src/Service.Host/client/src/components/manufacturingSkills/ManufacturingSkill.js
```diff
@@ -143,6 +145,37 @@ function ManufacturingSkill({
...
+                                    <Grid item xs={8}>
+                                        {!manufacturingSkill.dateInvalid && (
+                                            <Button
+                                                onClick={handleFieldChange(
+                                                    'dateInvalid',
+                                                    new Date()
+                                                )}
+                                            >
+                                                Make Invalid
+                                            </Button>
```
lewisrenfrew, line 174: doesnt really matter, but wouldnt just a date input field have been easier here (to let the user insert their own dateInvalid) rather than need both an inputfield and a button?

tests/Integration/Service.Tests/ManufacturingResourcesModuleSpecs/WhenGettingValid.cs
lewisrenfrew, line 1: not a comment on this file really but didnt know where else to put it - It doesn't look like there is an integration test similar to this one for the WhenGettingValidManufacturingSkills case

src/Service.Host/client/src/components/__tests__/manufacturingSkills.test.js
```diff
@@ -10,12 +10,14 @@ const manufacturingSkills = [
...
-        hourlyRate: 10
+        hourlyRate: 10,
+        dateInvalid: new Date().toISOString()
     },
     {
         skillCode: 'TESTCODE2',
         description: 'Descrip',
-        hourlyRate: 12
+        hourlyRate: 12,
+        dateInvalid: new Date().toISOString()
```
lewisrenfrew, line 20: same with all these other cases where you're marking every piece of test data as invalid. Seems an odd choice

src/Service/Modules/ManufacturingSkillsModule.cs
lewisrenfrew: I do realise that I said before to leave a GetAll endpoint in there too... what I really meant and should have said is that it felt like a good idea to keep supporting get all as well as filtering.... the best way to do that is with query params. If you do really want to keep two seperate endpoints for some reason then I would keep /manufacturing-skills to mean 'get all' and add a new /filter endpoint or something since /all feels ugly.... But really I think that  a single /manufacturing-resources that respects query params is by far the cleanest and most flexible solution

lewisrenfrew:
was deleting loads of dependabot branches and noticed this - Did you forget to merge?

